PR #30, opened 2019-02-27
update comment listener trigger format

### LINKED ISSUE #17: register() seems like it could be accidentally invoked
A bit of bike shedding but `register()` seems like it can occur in normal code people paste. Perhaps we should have something more name spaced to invoke Registrator?

tanmaykm:
Pattern accepted now: @JuliaRegistrator `register(...)`.

JuliaRegistrator is the GitHub account hosting the registrator app.

optional keyword args for `register`:
- branch: a branch to register (valid only on an issue comment)
- target: a specific registry to target (default is all registries configured with the app). e.g. `register(target=General)` or `register(target=Pro)`

fixes #17

### DIFF
--- a/src/Server.jl
+++ b/src/Server.jl
@@ -26,6 +26,7 @@ struct RegParams
     prid::Union{Nothing, Int}
     branch::Union{Nothing, String}
     comment_by_collaborator::Bool
+    target::Union{Nothing,String}
     cparams::CommonParams
 
     function RegParams(evt::WebhookEvent, phrase::RegexMatch)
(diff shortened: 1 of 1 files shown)


StefanKarpinski:
See https://github.com/JuliaComputing/Registrator.jl/issues/22 — should we change the name of that org before people get too used to the current name? I like the sound of putting "@JuliaRegistrar `register(...)`" in a comment since it sounds more like a person you're addressing.

tanmaykm:
:+1: @nkottary has access to the org I believe

PR #37, opened 2019-03-05
Approval automation

nkottary:


nkottary:
Need to get this merged: https://github.com/JuliaWeb/GitHub.jl/pull/132

tanmaykm:
may be good to avoid using a database if possible, in my opinion

nkottary:
I am using database to lookup the source PR/issue/commit from the registration PR. Should we do this some other way?

tanmaykm:
can we put the source reference in the PR text/comment and read it back when it is approved?

StefanKarpinski:
I had originally tried to make the PR on the registry predictable as "register/$package/v$version". Is this for looking that up or for going in the other direction back to the triggering pull request?

nkottary:
Its for going back to the triggering pull request. I'm adding some hidden text to the registry PR body which will be looked up for this purpose. Here's an example:

```
<!-- {
    "request_type": "issue",
    "pkg_repo_name": "nkottary/TestReg.jl",
    "trigger_id": "2",
    "tree_sha": "09b4cd50d5283670c1a698d9edb1fda1e7e606f2",
    "version": "0.1.0"
     } -->
```

I should also encrypt this JSON string so that it can't be edited and misused.

Issue #39, opened 2019-03-13
How can I use this for a non General registry?


davidanthoff:
We have our own registry for a bunch of lab internal stuff. Can I use Registrator for that use case? How would I configure it so that it works on our own registry, and not the General public one?

Is it general ready to be used?

StefanKarpinski:
@nkottary, any thoughts on this?

nkottary:
We support multiple registries. You have to say `register(target=your-registry-id)` where `your-registry-id` is configured in the [configuration file](https://github.com/JuliaComputing/Registrator.jl/blob/fbf683b884ad388e8e55d82beb43cd4145b9ac6d/image/scripts/sample.toml#L16)

StefanKarpinski:
@davidanthoff: please give it a try and file issues if there are any problems.

davidanthoff:
So that would require that we run a server somewhere?

tanmaykm:
Yes. Credentials to your private registry and repositories, that Registrator.jl needs to work with, would then remain on your server.

davidanthoff:
Hm, I actually would much prefer a setup where we could just use the normal registrator bot, and for example have a config file in the root of our packages where we can specify the URL of the registry for that specific package.

If the choice is between a) giving access to our private repos to the central registrator bot and b) running our own server, a) _easily_ wins out. Running our own infrastructure for this kind of thing is really kind of a no-starter for us...

davidanthoff:
I'm actually wondering whether https://github.com/JuliaLang/PkgDev.jl/pull/144 wouldn't be a perfectly fine workflow for our use-case...

nkottary:
It isn't necessary that you run your own server. Just install the normal registrator bot on both your private repo and the private registry. However, Registrator will not know which registry you are referring to when you just say `register()`. By default, it thinks `General`.

davidanthoff:
Is there a way to then tell it about the private registry?

nkottary:
We need to update the config on our end to whitelist your registry.

StefanKarpinski:
We can't really have the public registrator instance be an "open relay" to arbitrary private registries since that opens the server up to all sorts of mischief. The way this will have to work is that people can run their own registrator instances which are configured to run against their private registries.

BenJWard:
Is this simple to deploy on Heroku? I'd love to deploy this to manage the planned BioJulia registry

StefanKarpinski:
[...]

PR #65, opened 2019-04-08
Allow setting http_port through `$PORT`

BenJWard:
@nkottary As discussed in #39, this PR implements the Server getting its port from `ENV["PORT"]` if `http_port` is not set in the config.toml. If neither is set, then the server will use a default port of 8001.

This is useful for deploying places where you don't control the port number you get, like Heroku, which sets a value for `$PORT` when the app spins up.


nkottary:
Cool thanks!

PR #67, opened 2019-04-08
Make the server use the registry_deps option when calling the `register` function

BenJWard:
This PR adds an option to the config:

```
[registrator]
registry_deps = ["https://github.com/JuliaRegistries/General"]
```

That can be extended to add more registries that can be used for dependency checking. This is for situations where you are registering a package to your registry, which has a dependency in a different registry. Which might be frequently for non-General registries: It's likely non-General registries will have packages that have dependencies in General).

By default, General is already listed there in the sample config.


nkottary:
Thanks again!

PR #79, opened 2019-04-10
readme cleanup

joshday:


StefanKarpinski:
How about using a valid Markdown list format?

PR #81, opened 2019-04-10
Refactor and enable testing (prepare for more extensive testing)

iamed2:


src/Registrator.jl
```diff
@@ -3,14 +3,13 @@ module Registrator
...
 import Base: PkgId
-import Pkg: Pkg, TOML, GitTools
 
 const DEFAULT_REGISTRY = "https://github.com/JuliaRegistries/General"
 const REGISTRIES = Dict{String,UUID}()
 
 include("slack.jl")
 include("builtin_pkgs.jl")
-include("register.jl")
+include("regedit/RegEdit.jl")
```
ararslan:
```suggestion
include(joinpath("regedit", "RegEdit.jl"))
```

src/Registrator.jl
```diff
@@ -3,14 +3,13 @@ module Registrator
...
 import Base: PkgId
-import Pkg: Pkg, TOML, GitTools
 
 const DEFAULT_REGISTRY = "https://github.com/JuliaRegistries/General"
 const REGISTRIES = Dict{String,UUID}()
 
 include("slack.jl")
 include("builtin_pkgs.jl")
-include("register.jl")
+include(joinpath("regedit", "RegEdit.jl"))
```
fredrikekre:
```suggestion
include("regedit/RegEdit.jl")
```

StefanKarpinski:
@nkottary, would be willing to review this?

src/regedit/RegEdit.jl
```diff
@@ -0,0 +1,11 @@
+module RegEdit
```
fredrikekre: Internal extra modules like this is usually just more annoying than helpful IMO.

src/Registrator.jl
ararslan: What, why?

iamed2:
I'm planning to push commits as I make them; feel free to review in pieces or as a whole. This could also be merged piecemeal as commits will be atomic.

src/Registrator.jl
fredrikekre: Looks a million times nicer.

iamed2: I like it better with the slash as well; is it guaranteed to be valid though?

fredrikekre: Yes, https://github.com/JuliaLang/julia/pull/26656

src/regedit/RegEdit.jl
iamed2: In the end the rest of the code should be calling exactly one function in here (`register`). Having it in a module will help prevent leaky abstractions, and will enable branching it off into a separate package (which should probably happen eventually, given how different the functionality is between this and the server).

iamed2: I've definitely run into a few cases where it's annoying but I believe this will not be one of them.

src/Server.jl
```diff
@@ -12,7 +12,8 @@ using JSON
 using MbedTLS
 
 import Pkg: TOML
```
nkottary: This should be changed to just `import Pkg` no? I see that Server.jl uses `Pkg.Types.read_project` in a few places.

iamed2: Well it needs `TOML`, and there was already a `using Pkg` above

nkottary: Ah, you're right. Somehow I missed that.

iamed2:
Update: this is now just a PR for the reviewed commits; more will come in a new PR but this is good to merge as a first step.

PR #90, opened 2019-04-17
fix VersionSpec printing in Compat.toml

pfitzseb:
Fixes https://github.com/JuliaLang/Pkg.jl/issues/1155.

### DIFF
--- a/src/regedit/register.jl
+++ b/src/regedit/register.jl
@@ -288,7 +288,7 @@ function register(
         else
             compat_data = Dict()
         end
-        compat_data[pkg.version] = Dict{String,Any}(n=>Pkg.Types.semver_spec(v) for (n,v) in pkg.compat)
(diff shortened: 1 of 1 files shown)


StefanKarpinski:
A better fix here would be this:
```jl
compat_data[pkg.version] = Dict{String,Any}(
    n => let ranges = Pkg.Types.semver_spec(v).ranges
        length(ranges) == 1 ? string(ranges) : map(string, ranges)
    end for (n,v) in pkg.compat)
```
Untested.

StefanKarpinski:
Even better version that calls `compress_versions` correctly (hopefully—it's untested):
```jl
compat_data[pkg.version] = Dict{String,Any}(
    n => let spec = Pkg.Types.semver_spec(v)
        pool = # get vector of all versions of dependency
        ranges = compress_versions(pool, filter(in(spec), pool)).ranges
        length(ranges) == 1 ? string(ranges[1]) : map(string, ranges)
    end for (n,v) in pkg.compat)
```

PR #91, opened 2019-04-17
Clarify how backticks should be used

timholy:
Not sure if this is an improvement, but the previous README was ambiguous on this score.


nkottary:
Backticks are optional now. Also the generated tag and push commands are fixed now. So we can retain the bit about that.

timholy:
AFAICT we don't really need this then, so I'll close.

PR #99, opened 2019-04-22
added more clear instructions to the README file

chakravala:
based on this [discussion](https://discourse.julialang.org/t/ann-tagbot-creates-tags-and-releases-for-your-julia-packages-when-theyre-registered/23084/35?u=chakravala)

also, having a separate `full_readme.md` seems a bit non-sensical to me, especially since it wasn't even linked to anywhere... in my opinion that information should be merged to the main README

also, perhaps the info on this [web-page](https://juliaregistrator.github.io/) should also be synced with the other READMEs?


README.md
```diff
@@ -15,12 +15,18 @@ Registrator is a GitHub app that automates creation of registration pull request
 
 First, install the app on your package(s) as mentioned above.  The procedure for registering a new package is the same as for releasing a new version.
 
+1. Set the [Project.toml](Project.toml) version field in your repository to your new desired `version`.
```
StefanKarpinski: Maybe mention https://github.com/JuliaLang/Pkg.jl/blob/master/bin/gen_project.jl which can convert `REQUIRE` files to project files with (mostly) equivalent compat sections.

README.md
```diff
@@ -15,12 +15,18 @@ Registrator is a GitHub app that automates creation of registration pull request
...
+3. If something is incorrect, adjust, and redo step 2.
+4. Finally, either rely on [TagBot](https://github.com/apps/julia-tagbot) to tag and make a github release or alternatively tag the release manually.
+
 Registrator will look for the project file in the master branch by default, and will use the version set in the Project.toml file via, for example, `version = "0.1.0"`. To use a custom branch comment with:
 
 ```
 @JuliaRegistrator register(branch=name-of-your-branch)
 ```
 
+More detailed information about the usage of Registrator can be found in the [full readme](full_readme.md)
```
StefanKarpinski: How about just integrating the contents of `full_readme.md` into the main README? I agree that it seems weird to have two different README files.

StefanKarpinski:
Made some comments, but this is definitely a good clarification even as is.

chakravala:
Alright, I think I addressed all those points, but did not delete `full_readme.md`, I will leave that to you.

README.md
chakravala: I added the information in another section below

README.md
```diff
@@ -11,33 +11,72 @@ Registrator is a GitHub app that automates creation of registration pull request
 
 [![install](https://img.shields.io/badge/-install%20app-blue.svg)](https://github.com/apps/juliaregistrar/installations/new)
 
+#### Permissions and subscribed events for the app
```
StefanKarpinski: It's kind of unclear to me without additional context what this refers to.

vincentcp:
The full readme says in the last paragraph that the `Registrator.jl` can also be used for private registries, but the link is not working.

Can I still create registrator for my own registry, and, if so, where do I specify the location of my registry?

ianshmean:
Suggestion.. I think it would be helpful to have a section towards the start that says something like:

> Before starting it's good to check that your package conforms to the required Project.toml structure, and general package guidelines found [here](https://julialang.github.io/Pkg.jl/v1/creating-packages/)

README.md
nkottary: Those are instructions for someone who wants to host Registrator. There are 2 kinds of documentation here: 

1) For someone who has installed Registrator on their repo and wants to know how to register a new version.

2) For someone who wants to host Registrator and wants to know about the GitHub settings for the app. What to do for private packages etc.

I think 1) should go to README.md and 2) Should be in a file called docs.md (which is full_readme.md right now but I never got around to cleaning up/updating that file).

nkottary: Okay, I've removed this right now. Will add it later in a different PR.

vincentcp:
I follows the instructions on installing `JuliaRegistrar` app on my private package and adding the `JuliaRegistrator` collaborator to my private registry. How will the app decide that he has to send a PR to my private registry and not to `General.jl`. In the description of the app is stated 

> Creates registration pull requests to General.jl. 

So it seems that it will send a PR to `General.jl` instead

nkottary:
Please do not install Registrator on your private repo. There is no instruction about private repositories in the Readme: https://github.com/JuliaComputing/Registrator.jl/blob/master/README.md

We do not support private registrations in the current hosted instance of Registrator.

vincentcp:
There is in https://github.com/JuliaComputing/Registrator.jl/blob/master/full_readme.md, hence the confusion.

nkottary:
Okay, I've updated the readme in this PR regarding private packages.

vincentcp:
Thank you ! This contains the information I need.

PR #106, opened 2019-04-23
Remove use of some Base functions in check_version.

fredrikekre:
Those functions does not belong in Base IMO, so better not use them and we can remove them from there (https://github.com/JuliaLang/julia/compare/cb85f840a533f79741b79dfd94a00d9581f22983..2e0d6b75356f101ee7f743bbb74cf7d6ce4f8a79)


StefanKarpinski:
Agree, we should delete this function from Base as well once all uses of it here are removed.

PR #109, opened 2019-04-23
Update PR body when it already exists

### LINKED ISSUE #108: Update commit hash when register() is called on a new commit
See https://github.com/christopher-dG/tag-bot/issues/3 for tag-bot issue.

Because the commit hash is not updated when register() is called to fix a commit, tag-bot does not tag the correct commit.

christopher-dG:
Fixes #108 
Fixes https://github.com/christopher-dG/tag-bot/issues/3 (am I allowed to do this?!)

I haven't tested this, maybe someone more familiar with test-deploying Registrator and using GitHub.jl can tell me if this is going to break or not.

### DIFF
--- a/src/Server.jl
+++ b/src/Server.jl
@@ -736,13 +736,24 @@ function make_pull_request(pp::ProcessedParams, rp::RequestParams, rbrn::RegBran
     pr = nothing
     repo = join(split(target_registry["repo"], "/")[end-1:end], "/")
     msg = ""
+    auth = get_user_auth()
     try
-        pr = create_pull_request(repo; auth=get_user_auth(), params=params)
+        pr = create_pull_request(repo; auth=auth, params=params)
         msg = "created"
         @debug("Pull request created")
     catch ex
         if is_pr_exists_exception(ex)
-            @debug("Pull request already exists, not creating")
+            @debug("Pull request already exists, trying to update body")
+            params = Dict(
+                "base" => target_registry["base_branch"],
+                "head" => split(repo, "/")[1] * ":" * brn,
+            )
+            prs = pull_request(repo; auth=auth, params=params)
+            if !isempty(params)
+                pr = prs[1]
+                params = Dict("body" => params["body"])
+                update_pull_request(repo, pr; auth=auth, params=params)
+            end
             msg = "updated"
         else
             rethrow(ex)
@@ -752,7 +763,6 @@ function make_pull_request(pp::ProcessedParams, rp::RequestParams, rbrn::RegBran
     if pr == nothing
         # Look for pull request in last 10 pages, each page contains 15 PRs
         params = Dict("state"=>"open", "per_page"=>15)
-        auth = get_user_auth()
         prs, page_data = pull_requests(repo; auth=auth, params=params, page_limit=1)
         for p in prs
             if p.base.ref == target_registry["base_branch"] && p.head.ref == brn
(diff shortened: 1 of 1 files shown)


src/Server.jl
```diff
@@ -743,6 +743,8 @@ function make_pull_request(pp::ProcessedParams, rp::RequestParams, rbrn::RegBran
     catch ex
         if is_pr_exists_exception(ex)
             @debug("Pull request already exists, not creating")
+            params = Dict("body" => params["body"])
```
fredrikekre: You need more params here, in particular the PR number since the call on the next line wouldn't know what PR to update.

christopher-dG: Oh right, of course.

christopher-dG: I think this might do the right thing now

src/Server.jl
```diff
@@ -736,13 +736,24 @@ function make_pull_request(pp::ProcessedParams, rp::RequestParams, rbrn::RegBran
...
     try
-        pr = create_pull_request(repo; auth=get_user_auth(), params=params)
+        pr = create_pull_request(repo; auth=auth, params=params)
         msg = "created"
         @debug("Pull request created")
     catch ex
         if is_pr_exists_exception(ex)
-            @debug("Pull request already exists, not creating")
+            @debug("Pull request already exists, trying to update body")
+            params = Dict(
```
fredrikekre: Do you need to create this? Can't you just use the one from L715?

src/Server.jl
```diff
@@ -736,13 +736,24 @@ function make_pull_request(pp::ProcessedParams, rp::RequestParams, rbrn::RegBran
...
+                "base" => target_registry["base_branch"],
+                "head" => split(repo, "/")[1] * ":" * brn,
+            )
+            prs = pull_request(repo; auth=auth, params=params)
+            if !isempty(prs)
+                pr = prs[1]
+                params = Dict("body" => params["body"])
```
fredrikekre: You have overwritten `params` here on L747 so the body does not exist, right?

src/Server.jl
christopher-dG: Well I don't want to send any unwanted values that may or may not affect the API call.

src/Server.jl
christopher-dG: Right, that's true

christopher-dG:
I've dealt with the overwritten variable by just constructing the parameters inline.

src/Server.jl
```diff
@@ -736,13 +736,22 @@ function make_pull_request(pp::ProcessedParams, rp::RequestParams, rbrn::RegBran
...
-        pr = create_pull_request(repo; auth=get_user_auth(), params=params)
+        pr = create_pull_request(repo; auth=auth, params=params)
         msg = "created"
         @debug("Pull request created")
     catch ex
         if is_pr_exists_exception(ex)
-            @debug("Pull request already exists, not creating")
+            @debug("Pull request already exists, trying to update body")
+            prs = pull_request(repo; auth=auth, params=Dict(
+                "base" => target_registry["base_branch"],
```
fredrikekre:
```suggestion
                "base" => params["base"],
```

src/Server.jl
```diff
@@ -736,13 +736,22 @@ function make_pull_request(pp::ProcessedParams, rp::RequestParams, rbrn::RegBran
...
+                "head" => split(repo, "/")[1] * ":" * brn,
```
fredrikekre:
```suggestion
                "head" => params["head"],
```

christopher-dG: I don't think this does what we want, the GitHub docs seem to say that the value must also contain the user. https://developer.github.com/v3/pulls/#parameters

PR #111, opened 2019-04-23
Check [extras] when determining version compatibility

### LINKED ISSUE #110: Test-only dependencies in [compat] cause an error
See https://github.com/JuliaStats/Distributions.jl/issues/868#issuecomment-485991894. Distributions has a test dependency on JSON, and lists it in `[compat]` and `[extras]`. The fact that it isn't in `[deps]` makes the registration break.

ararslan:
Fixes #110.

### DIFF
--- a/src/regedit/register.jl
+++ b/src/regedit/register.jl
@@ -105,7 +105,7 @@ function findpackageerror(name::String, u::String, regdata::Array{RegistryData})
         if haskey(_registry_data.packages, u)
             name_in_reg = _registry_data.packages[u]["name"]
             if name_in_reg != name
-                return "Error in `[deps]`: UUID $u refers to package '$name_in_reg' in registry but deps file has '$name'"
+                return "Error in Project.toml: UUID $u refers to package '$name_in_reg' in registry but Project.toml has '$name'"
             end
             return nothing
         end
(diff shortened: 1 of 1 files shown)


ararslan:
I have no idea whatsoever how to test this or how most of Registrator works, so some help and guidance would be appreciated.

iamed2:
You'll want to add a check like https://github.com/JuliaComputing/Registrator.jl/blob/45d066138b2e9a3939a8c01f2aaef47371c2c584/src/regedit/register.jl#L262 for extras (just checks to see that they're available in the registries). 

Then you'll also want to run the `findpackageerror` part of the `else` clause right below what you've added.

IMO all the targets should be checked for version compatibility but this is fine for now.

ararslan:
Is this more like what you have in mind, @iamed2?

src/regedit/register.jl
```diff
@@ -348,18 +348,29 @@ function register(
...
-                uuidofdep = string(pkg.deps[n])
 
                 err = findpackageerror(n, uuidofdep, regdata)
                 if err !== nothing
                     @debug(err)
                     return RegBranch(pkg, branch; er=err)
                 end
+
+                if inextras && !indeps
+                    @debug("$n is a test-only dependency; omitting from Versions.toml")
```
iamed2:
```suggestion
                    @debug("$n is a test-only dependency; omitting from Compat.toml")
```

iamed2:
This is good, just one suggestion because this section reads Versions.toml but writes Compat.toml.

PR #117, opened 2019-04-25
More efficient lookup for existing pull request

christopher-dG:
Instead of looping over *all* pull requests, we can tell GitHub which one we want by specifying the head and base branches. There cannot be multiple open PRs with identical head and base branches, so we get at most one result.

Also did a quick find and replace over the file to "properly" compare to `nothing`.

### DIFF
--- a/src/Server.jl
+++ b/src/Server.jl
@@ -762,32 +762,14 @@ function make_pull_request(pp::ProcessedParams, rp::RequestParams, rbrn::RegBran
     end
 
     if pr == nothing
-        # Look for pull request in last 10 pages, each page contains 15 PRs
-        prs, page_data = pull_requests(repo; auth=auth,
-                                       params=Dict("state"=>"open", "per_page"=>15),
-                                       page_limit=1)
-        for p in prs
-            if p.base.ref == target_registry["base_branch"] && p.head.ref == brn
-                @debug("PR found")
-                pr = p
-                break
-            end
-        end
-
-        if pr == nothing
-            i = 1
-            while i < 10
-                prs, page_data = pull_requests(repo; auth=auth, page_limit=1, start_page=page_data["next"]);
-                for p in prs
-                    if p.base.ref == target_registry["base_branch"] && p.head.ref == brn
-                        @debug("PR found")
-                        pr = p
-                        break
-                    end
-                end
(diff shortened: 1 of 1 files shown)


christopher-dG:
Man codecov on largely untested projects is so demoralizing 🙃

src/Server.jl
```diff
@@ -761,36 +761,18 @@ function make_pull_request(pp::ProcessedParams, rp::RequestParams, rbrn::RegBran
...
-            end
+    if pr === nothing
+        prs, _ = pull_requests(repo; auth=auth, params=Dict(
+            "state" => "open",
+            "base" => params["base"],
+            "head" => string(split(repo, "/")[1], ":", params["head"]),
+        ))
+        if !isempty(prs)
+            @debug("PR found")
+            pr = prs[1]
```
fredrikekre: Maybe add an `@assert length(prs) == 1`?

nkottary:
Is `isnothing` a better way to check nothingness?

christopher-dG:
`isnothing` is Julia 1.1 and up, I'd rather not depend on that.

nkottary:
I tried out the `pull_requests` call from a terminal. It does not seem to filter by `head` and `base` that we pass as `params`. It just gives me all the PRs.

christopher-dG:
Did you prepend the head with the user name and a colon? I did the same thing manually.

nkottary:
Like this?

```
julia> a, b = pull_requests("JuliaRegistries/General"; params=Dict("state" => "open","base" => "JuliaComputing:master", "head" => "JuliaComputing:register/PowerDynBase/v1.0.0",))
(PullRequest[], Dict{String,String}())

julia> a
0-element Array{PullRequest,1}```

**Comment by christopher-dG** [comment]
`base` is just master, since it can't be in any other repo.

**Comment by fredrikekre** [comment]
```
julia> a, b = GitHub.pull_requests("JuliaRegistries/General", params=Dict("state"=>"open", "base"=>"master", "head" => "JuliaRegistries/General:register/CapacityExpansion/v0.1.1"))
(PullRequest[PullRequest(324)], Dict{String,String}())
```

nkottary:
Ok, thanks.

PR #120, opened 2019-04-27
Enable organizational membership check for release rights

ianshmean:
This should fix the inability of *Members* (as opposed to *Owners*) of org-owned repos to invoke `@JuliaRegistrator`, even if they have release rights.

Designed to handle both repos owned by organizations through membership, and previous method for collaborators on privately owned repos

**N.B.** I haven't tested this code locally, as I was unsure how to on my local machine

### DIFF
--- a/src/Server.jl
+++ b/src/Server.jl
@@ -54,6 +54,7 @@ struct RequestParams{T<:RequestTrigger}
 
     function RequestParams(evt::WebhookEvent, phrase::RegexMatch)
         reponame = evt.repository.full_name
+        user = get_user_login(evt.payload)
         trigger_src = EmptyTrigger()
         commenter_can_release = false
         err = nothing
@@ -91,6 +92,12 @@ struct RequestParams{T<:RequestTrigger}
                 else
                     err = "Commenter doesn't have release rights"
                     @debug(err)
+                    if is_owned_by_organization(evt)
+                        org = evt.repository.owner.login
+                        make_comment(evt, "**Register Failed**\n@$(user), it looks like you are not a publicly listed member/owner in the parent organization ($(org)).\nIf you are a member/owner, you will need to change your membership to public. See [GitHub Help](https://help.github.com/en/articles/publicizing-or-hiding-organization-membership)")
+                    else
+                        make_comment(evt, "**Register Failed**\n@$(user), it looks like you don't have collaborator status on this repository.")
+                    end
                 end
                 @debug("Comment is on a pull request")
             else
--- a/src/Server.jl
+++ b/src/Server.jl
@@ -412,13 +412,7 @@ function is_comment_by_org_owner_or_member(event)
     return any(userorgs.==org)
 end
 
-function has_release_rights(event)
-    if is_comment_by_collaborator(event)
-        return true
-    else
-        return is_owned_by_organization(event) && is_comment_by_org_owner_or_member(event)
-    end
-end
+has_release_rights(event) = is_comment_by_collaborator(event) || is_owned_by_organization(event) && is_comment_by_org_owner_or_member(event)
 
 function is_pull_request(payload)
     haskey(payload, "pull_request") || haskey(payload, "issue") && haskey(payload["issue"], "pull_request")
--- a/src/Server.jl
+++ b/src/Server.jl
@@ -408,8 +408,11 @@ function is_comment_by_org_owner_or_member(event)
     @debug("Checking if comment is by repository parent organization owner or member")
     org = event.repository.owner.login
     user = get_user_login(event.payload)
-    userorgs = orgs(user)
-    return any(userorgs.==org)
+    if get(config["registrator"], "check_private_membership", false)
+        return GitHub.check_membership(org, user; auth=get_user_auth())
+    else
+        return GitHub.check_membership(org, user; public_only=true)
+    end
 end
 
 has_release_rights(event) = is_comment_by_collaborator(event) || is_owned_by_organization(event) && is_comment_by_org_owner_or_member(event)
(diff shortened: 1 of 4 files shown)


StefanKarpinski:
Can you say a bit about the desired permission model here? Does this make it match users who can create a release on GitHub? If so, that does seem like the right permission to match.

ianshmean:
The general idea is:
```
-If repo is owned by organization 
--Check if commenter is member of organization via checking repo org against orgs(user)
-else
--Check if commenter is collaborator (previous method)
```

ianshmean:
A direct check for release rights would probably be cleaner, but I couldn't find that method in GitHub.jl or the github api. But I may need to look deeper

christopher-dG:
There's no such thing as release rights so don't look too hard. Also, I believe there's a "check org membership" endpoint so we should probably use that if GitHub.jl exposes it.

src/Server.jl
```diff
@@ -382,12 +382,32 @@ function get_sha_from_branch(reponame, brn; auth = GitHub.AnonymousAuth())
...
     @debug("Checking if comment is by collaborator")
     user = get_user_login(event.payload)
     return iscollaborator(event.repository, user; auth=get_access_token(event))
 end
 
+function is_comment_by_org_owner_or_member(event)
+    @debug("Checking if comment is by repository parent organization owner or member")
+    org = event.repository.owner.login
+    user = get_user_login(event.payload)
+    userorgs = orgs(user)
```
nkottary:
```
julia> orgs("nkottary")
(Owner[Owner("JuliaDatabases")], Dict{String,String}())

```

I am an owner/member in 3 other orgs, this just returns one. I tried it with an `auth` param too. Didn't help.

src/Server.jl
```diff
@@ -382,12 +382,32 @@ function get_sha_from_branch(reponame, brn; auth = GitHub.AnonymousAuth())
...
+    return any(userorgs.==org)
+end
+
+function has_release_rights(event)
+  if is_owned_by_organization(event)
+    return is_comment_by_org_owner_or_member(event)
+  else
+    return is_comment_by_collaborator(event)
+  end
```
nkottary: The condition should be:

```
if is_comment_by_collaborator(event)
    return true
else
    return is_owned_by_organization(event) && is_comment_by_org_owner_or_member(event)
end
```

ianshmean:
> I believe there's a "check org membership" endpoint so we should probably use that if GitHub.jl exposes it.

@christopher-dG `GitHub.check_membership("JuliaIO","ianshmean",auth=myauth)` works but requires `repo  Full control of private repositories` from the oauth permissions. Does Registrator require that permission?

simonbyrne:
Creating releases requires write permission to the repo (which is why tagbot requires it). We should allow anyone with write permission to the repo to create a release, which members may or may not have

simonbyrne:
I think this is what should be used: https://developer.github.com/v3/repos/collaborators/#review-a-users-permission-level

ianshmean:
I've implemented `collaboratorpermission(repo,user)` here https://github.com/JuliaWeb/GitHub.jl/pull/133 

But I'm now confused because I'm getting `true` for `GitHub.iscollaborator()` on other org repos that I've never interacted with, while being *only* a member, so perhaps I'm barking up the wrong tree as to the reason for not being able to invoke `@JuliaRegistrator register()`

Irrespective, checking `collaboratorpermission(repo,user)` is a good idea as @simonbyrne points out

nkottary:
No need to check `collaboratorpermission`.

Using `GitHub.check_membership(org, user; public_only=true)` is fine. For things to work the membership of a user to an Org must be public. This does not require any private repo permission.

ianshmean:
Ok, although the problem with the `GitHub.check_membership()` approach is precisely that very few people make their org membership public, as it defaults to private. I think there would still be a fair amount of confusion as to why it's not working. Perhaps there should be an error warning for people with insufficient rights, rather than the current lack of message? Is that possible?

As I mentioned above, I am a little confused now though.. I'm still only a *Member* of JuliaIO, and `JuliaIO/ImageMagick` is not listed in https://github.com/settings/repositories which is where any repos I'm a collaborator on should be listed (I am a collaborator on `JuliaIO/VideoIO` and that appears)

Yet `GitHub.iscollaborator("JuliaIO/ImageMagick.jl","ianshmean",auth=myauth)` = `true`

*Collaborator* status seems to be complicated...

christopher-dG:
IMO it's not unreasonable that to register a package, your membership should be public. But others may disagree. 
And there should absolutely be an error message for all failed registrations.

christopher-dG:
As far as collaborator/member goes, this is the extent of my knowledge:

- if you have push access to another _individual's_ repo, you are a collaborator
- if you have push access to an _organisation_'s repo, you are either a member of that org or a collaborator.

So my opinion is still this:

- for org-owned repos, check membership of org
- for individuals' repos, check collaborator status

With a single addition:

- for org-owned repos, check collaborator status if they are not an organisation member

ianshmean:
FWIW I think that's fair, as long as there was a friendly warning. Something like:

For privately owned:

----

**Register Failed**
It looks like you don't have collaborator status on this repository.

----

For org owned:

----

**Register Failed**
It looks like you are not a publicly listed member/owner in the parent organization ($(org)).
If you are a member/owner, you will need to change your membership to public. [GitHub Help](https://help.github.com/en/articles/publicizing-or-hiding-organization-membership)

----

StefanKarpinski:
I like requiring people to make their membership in an org public. If they're not willing to do that, then we can't know that they're allowed to make releases. I also think that we want to encourage membership in Julia orgs to be public for multiple reasons, including that we want it to be visible that there are many members and that we want who those members are to be knowable to people outside of the org.

ianshmean:
I agree 👍🏻I've added those failure comments, with the addition of tagging the commenter to help highlight the error. (Is the markdown formatting right for the comments?)

And again, PR is untested at my end due to need for github integration

PR #121, opened 2019-04-29
Fix displaying Registrator commit when it is installed with Pkg.add

### LINKED ISSUE #118: Fix issues with getting Registrator commit
https://github.com/JuliaComputing/Registrator.jl/commit/171d0722222d9bca6352852c811f8f5074393643#commitcomment-33321721

nkottary:
Fix #118

### DIFF
--- a/src/regedit/register.jl
+++ b/src/regedit/register.jl
@@ -417,7 +417,9 @@ function register(
 
         Pkg.Compress.save(compat_file, compat_data)
 
-        reg_pkgs = filter(x -> x.name == "Registrator", Pkg.Display.status(Pkg.Types.Context()))
+        reg_pkgs = Pkg.Display.status(Pkg.Types.Context(),
+                                      [PackageSpec("Registrator",
+                                                   Base.UUID("4418983a-e44d-11e8-3aec-9789530b3b3e"))])
         if length(reg_pkgs) == 0
             reg_commit = "unknown"
         else
--- a/src/regedit/register.jl
+++ b/src/regedit/register.jl
@@ -417,6 +417,16 @@ function register(
 
         Pkg.Compress.save(compat_file, compat_data)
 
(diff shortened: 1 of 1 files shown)


src/regedit/register.jl
```diff
@@ -417,6 +417,16 @@ function register(
 
         Pkg.Compress.save(compat_file, compat_data)
 
+        reg_pkgs = filter(x -> x.name == "Registrator", Pkg.Display.status(Pkg.Types.Context()))
```
nkottary: Is there a better way to do this?

src/regedit/register.jl
```diff
@@ -417,6 +417,18 @@ function register(
 
         Pkg.Compress.save(compat_file, compat_data)
 
+        reg_pkgs = Pkg.Display.status(Pkg.Types.Context(),
+                                      [PackageSpec("Registrator",
+                                                   Base.UUID("4418983a-e44d-11e8-3aec-9789530b3b3e"))])
+        if length(reg_pkgs) == 0
+            reg_commit = "unknown"
+        else
+            reg_commit = reg_pkgs[1].new.hash
```
fredrikekre: This is not a commit hash, its a tree hash.

src/regedit/register.jl
```diff
@@ -417,6 +417,18 @@ function register(
...
+            if reg_commit === nothing    # Registrator is dev'd
+                reg_commit = LibGit2.head(Pkg.dir("Registrator"))
```
fredrikekre: `Pkg.dir` should not be used, use `pathof(Registrator)` instead.

src/regedit/register.jl
nkottary: How can I get the commit hash?

fredrikekre: Don't know, I don't think there is an easy way. Why not just use the version instead when `Registrator` is added?

PR #131, opened 2019-05-05
Remove restriction that repos must end with .jl.

fredrikekre:
fixes https://github.com/fredrikekre/jlpkg/commit/4797913f41002874c3b4649e8f83e17792de9e8c#commitcomment-33418965

Edit: better view of diff: https://github.com/JuliaRegistries/Registrator.jl/pull/131/files?w=1

### DIFF
--- a/src/Server.jl
+++ b/src/Server.jl
@@ -77,38 +77,32 @@ struct RequestParams{T<:RequestTrigger}
             err = "Private registration request recieved, ignoring"
             @debug(err)
         elseif action_name == "register"
-            if endswith(reponame, ".jl")
-                commenter_can_register = has_register_rights(evt)
-                if commenter_can_register
-                    @debug("Commenter has registration rights")
-                    if is_pull_request(evt.payload)
-                        if config["registrator"]["disable_pull_request_trigger"]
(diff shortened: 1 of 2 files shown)


fredrikekre:
I re-enabled Travis on this repo after the move to this org. Looks like it fails for the new WebUI stuff @christopher-dG

nkottary:
I think we should disable the WebUI tests for travis. It requires secret keys which we can't make public.

fredrikekre:
You can add secret keys as environment variables on travis.

nkottary:
Okay, so I've added `GITHUB_API_TOKEN` env variable in travis settings. @fredrikekre feel free to close and reopen the PR to re-run travis.

fredrikekre:
Seems like something else is wrong then?

fredrikekre:
Disabled the failing test for now.

fredrikekre:
Do we prefer a warning? E.g.
```
The repository name does not end with '.jl', which is a convention used for most Julia packages.
```

nkottary:
This change is deployed.

nkottary:
I think a warning would be good to have. Can you open an issue to discuss whether there needs to be a warning?

fredrikekre:
Will do.

PR #133, opened 2019-05-06
Fix failing web test, update repo links

christopher-dG:
Turns out we should use repositories that we know will continue to exist as test examples :upside_down_face:. 
A bit of relevant discussion in #131.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,7 +1,7 @@
 # Registrator
 
-[![Build Status](https://travis-ci.com/JuliaComputing/Registrator.jl.svg?branch=master)](https://travis-ci.com/JuliaComputing/Registrator.jl)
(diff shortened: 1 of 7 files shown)


christopher-dG:
Ah, looks like the GitHub token is not available for pull request builds. Perhaps we should disable this set of tests which requires the GitHub API for pull requests?

fredrikekre:
IIRC they are available for PRs with head in the same repository, so it worked for e.g. #131 but this PR seems to be from your fork. Maybe we can just disable them if `get(ENV, "GITHUB_API_TOKEN", nothing) === nothing` or something?

christopher-dG:
Ah, that make sense. I'll do that check on top of checking for the travis variable, since testing locally without an API key works fine.

christopher-dG:
I have also replaced the links to JuliaComputing with JuliaRegistries.

nkottary:
Thank you, @christopher-dG

Issue #130, opened 2019-05-03
Rethink the trigger command


nkottary:
Currently we use a some what julia syntax to express keyword arguments to the trigger. Ex.:

```
@JuliaRegistrator register(target="SomeRegistry", branch="release")
```

Which seems too complex. Maybe a simpler alternative would be to mention the arguments line by line:

```
@JuliaRegistrator register()

/branch release
/target MyRegistry
```

Thoughts?

iamed2:
What is too complex?

nkottary:
Using julia syntax :)

christopher-dG:
IMO there's no great syntax that we can achieve with a GitHub comment, but I do think that using pretend code is awkward. I'd prefer your proposed syntax, although I'd get rid of the `()` as well. 
But I don't think that's really worth triggering the annoyance of everyone by yet another breaking workflow change. 
If only we could use an interface more specialized for filling out and submitting forms :stuck_out_tongue_winking_eye:.

oxinabox:
Triggering the parser seems like it is exposing a lot of code that has not been security hardenned to the wild world.

christopher-dG:
Yeah, if we do want to keep existing syntax to avoid annoying people, it would be easy to parse via a regex.

christopher-dG:
I'll close this for now since I think the new parser is acceptable, feel free to reopen if there's more to say.

PR #135, opened 2019-05-07
Decouple server and registrator

nkottary:
This PR lets us run the GitHub comment listener and Registrator (the thing that does `register`) as separate processes. So we will be able to restart Registrator without any downtime (The server will queue the requests while registrator is down). This is useful when deploying fixes to `register`.

TODO: Do the same for the web UI.


nkottary:
This works now!

#### This PR makes Registrator run as three julia processes:
1) The registration service which is the backend
2) Comment bot
3) Web UI

Also a bunch of work was done to move code into smaller files.

#### Changes to WebUI:
1) User requests are queued and are sent one by one to the backend. Earlier we could have multiple register()'s happen concurrently which is not correct.
2) I've added targets to stop, start and restart WebUI to the makefile.

#### Non-critical pending issues:

Will fix these up in future PRs
1) Mux does not shutdown gracefully when socket is closed like it says [here](https://github.com/JuliaWeb/HTTP.jl/blob/2f65e07e2166991b0b78be466aa282cdf26b7161/src/Servers.jl#L141).
2) Polling of ZMQ socket is inefficient right now. I've left a note [here](https://github.com/JuliaRegistries/Registrator.jl/blob/17217839a9b0a995a0d14e3ffaf26a3c06d2db0c/src/Messaging.jl#L131)

@christopher-dG Let me know if the WebUI changes look good to you.

docs.md
```diff
@@ -4,23 +4,33 @@ This document describes how you can host your own instance of Registrator.
...
-```
-julia -e 'using Registrator; Registrator.RegServer.main()' conf.toml
-```
+You can use `make` to control the components of registrator. The `Makefile` can be found in the `run` directory. Commands:
 
-To safely stop the Registrator server, `touch` the file mentioned as `stop_file` in the config. By default it is `/tmp/stopregistrator`:
+1) `make startall`
+2) `make startregservice`
+3) `make startcommentbot`
+4) `make startwebui`
```
StefanKarpinski: Maybe use `start-regservice` and `start-commentbot`, etc.?

src/webui/WebUI.jl
```diff
@@ -0,0 +1,227 @@
...
+
+# U is a User type, e.g. GitHub.User.
+struct User{U, F <: GitForge.Forge}
+    user::U
+    forge::F
+end
+
+struct RegistrationData
+    project::Pkg.Types.Project
+    tree::AbstractString
```
christopher-dG: The `AbstractString`s in these structs should all be `String`s.

src/webui/WebUI.jl
```diff
@@ -0,0 +1,227 @@
...
+
+    @info "Starting WebUI"
+    monitor = @async status_monitor(CONFIG["stop_file"], event_queue, httpsock)
+    reqproc = @async request_processor(zsock)
+    start_server(ip, port)
+    wait(reqproc)
+    wait(monitor)
+
+    # The !stopped! part is grep'ed by restart.sh, do not change
+    @info "Server !stoppped!"
```
christopher-dG:
```suggestion
    @info "Server !stopped!"
```

src/webui/WebUI.jl
```diff
@@ -0,0 +1,227 @@
...
+        r -> html(404, "Page not found"),
+    )
+    do_action() = begin; t = serve(server, ip, port; server=httpsock[], readtimeout=0); wait(t); end
+    handle_exception(ex) = (isa(ex, Base.IOError) && (ex.code == -103)) ? :exit : :continue
+    keep_running() = isopen(httpsock[])
+    @info "Serving" ip port
+    recover("webui", keep_running, do_action, handle_exception)
+end
+
+function main(config::AbstractString="config.toml")
```
christopher-dG: This should probably be something along the lines of:

```jl
function main(config::AbstractString=isempty(ARGS) ? "config.toml" : first(ARGS))
```
It's most intuitive if the server dies immediately for non-existent files.

src/webui/WebUI.jl
```diff
@@ -0,0 +1,227 @@
...
+    @error "Handler error" route=r.target
+    html(500, "Server error, sorry!")
+end
+
+pathmatch(p::AbstractString, f::Function) = branch(r -> first(split(r.target, "?")) == p, f)
+
+function action(regdata::RegistrationData, zsock::RequestSocket)
+    regp = RegEdit.RegisterParams(cloneurl(regdata.repo), regdata.project, regdata.tree;
+        registry=REGISTRY[].clone, registry_deps=REGISTRY[].deps, push=true,
+        gitconfig=Dict("user.name"=>CONFIG["user"], "user.email"=>CONFIG["email"]),
```
christopher-dG:
```suggestion
        gitconfig=Dict("user.name" => CONFIG["user"], "user.email" => CONFIG["email"]),
```

src/webui/WebUI.jl
```diff
@@ -0,0 +1,227 @@
...
+        end
+    end
+    @debug msg
+    REGISTRATIONS[regdata.commit] = RegistrationState(msg, state)
+    nothing
+end
+
+function request_processor(zsock::RequestSocket)
+    do_action() = action(take!(event_queue), zsock)
+    handle_exception(ex) = (isa(ex, InvalidStateException) && (ex.state == :closed)) ? :exit : :continue
```
christopher-dG:
```suggestion
    handle_exception(ex) = ex isa InvalidStateException && ex.state === :closed ? :exit : :continue
```

src/webui/WebUI.jl
```diff
@@ -0,0 +1,227 @@
...
+        pathmatch(ROUTE_INDEX, index),
+        pathmatch(ROUTE_AUTH, auth),
+        pathmatch(ROUTE_CALLBACK, callback),
+        pathmatch(ROUTE_SELECT, select),
+        pathmatch(ROUTE_REGISTER, register),
+        pathmatch(ROUTE_STATUS, status),
+        r -> html(404, "Page not found"),
+    )
+    do_action() = begin; t = serve(server, ip, port; server=httpsock[], readtimeout=0); wait(t); end
+    handle_exception(ex) = (isa(ex, Base.IOError) && (ex.code == -103)) ? :exit : :continue
```
christopher-dG:
```suggestion
    handle_exception(ex) = ex isa Base.IOError && ex.code == -103 ? :exit : :continue
```

src/webui/WebUI.jl
nkottary: Thanks! Good catch.

StefanKarpinski:
Please be sure to squash when merging! (Or clean up the commit history before merging.)

src/webui/WebUI.jl
christopher-dG: Also, in my opinion these values should not be required, since global Git configuration can be used instead.
Perhaps:
```julia
gitconfig = Dict{String, String}()
haskey(CONFIG, "user") && (gitconfig["user.name"] = CONFIG["user"])
haskey(CONFIG, "email") && (gitconfig["user.email"] = CONFIG["email"])
```

src/webui/routes/select.jl
```diff
@@ -0,0 +1,87 @@
+# Step 4: Select a package.
+function select(::HTTP.Request)
```
christopher-dG: It would be really nice if this application did not require JavaScript. This could be reworked to happen in the backend.

src/webui/routes/status.jl
```diff
@@ -0,0 +1,18 @@
...
+function status(r::HTTP.Request)
+    r.method == "GET" || return json(405, Dict("error" => "Method not allowed"))
+
+    # Only allow authenticated users
+    state = getcookie(r, "state")
+    if isempty(state) || !haskey(USERS, state)
+        return json(400, Dict("error" => "Missing or invalid state cookie"))
+    end
+
+    uri = HTTP.URI(r.target)
```
christopher-dG:
```jl
id = getquery(r, "id")
```

src/webui/routes/status.jl
```diff
@@ -0,0 +1,18 @@
...
+    params = parseform(uri.query)
+    id = params["id"]
+    !haskey(REGISTRATIONS, id) && return json(404, Dict("state"=>"unknown"))
+    regstate = REGISTRATIONS[id]
+    regstate.state == :pending || delete!(REGISTRATIONS, id)
```
christopher-dG:
```suggestion
    regstate.state === :pending || delete!(REGISTRATIONS, id)
```

src/webui/routes/status.jl
```diff
@@ -0,0 +1,18 @@
...
+    # Only allow authenticated users
+    state = getcookie(r, "state")
+    if isempty(state) || !haskey(USERS, state)
+        return json(400, Dict("error" => "Missing or invalid state cookie"))
+    end
+
+    uri = HTTP.URI(r.target)
+    params = parseform(uri.query)
+    id = params["id"]
+    !haskey(REGISTRATIONS, id) && return json(404, Dict("state"=>"unknown"))
```
christopher-dG:
```suggestion
    haskey(REGISTRATIONS, id) || return json(404, Dict("state"=>"unknown"))
```

src/webui/webutils.jl
```diff
@@ -0,0 +1,64 @@
...
+            <br>
+            $body
+          </body>
+        </html>
+        """
+    return HTTP.Response(status, ["Content-Type" => "text/html"]; body=doc)
+end
+
+json(body::Dict) = json(200, body)
+json(status::Int, body::Dict) = HTTP.Response(status, ["Content-Type" => "text/json"]; body=JSON.json(body))
```
christopher-dG: This would be easier to use if it were:
```jl
json(status::Int; kwargs...) = HTTP.Response(#=...=#; body=JSON.json(kwargs))
```

christopher-dG:
Have we considered just cloning the target registry to a unique, clean directory for each registration? That would make concurrency a lot simpler. Dealing with DDOS etc. would just be checking if the parent directory has N or more registry clones and if so, saying "please come back in a lil bit".

run/commentbot.sh
```diff
@@ -0,0 +1,3 @@
+#!/bin/bash
```
oxinabox:
```suggestion
#!/bin/sh
```

run/config.toml
```diff
@@ -1,8 +1,5 @@
-[server]
-# Remove http_port to make the server use ENV["PORT"]. Heroku users must do this.
-http_port = 8001
-http_ip = "0.0.0.0"
-stop_file = "/tmp/stopregistrator"    # `touch` this file to stop Registrator
+[regservice]
+stop_file = "/tmp/stopregservice"    # `touch` this file to stop registration service
```
oxinabox:
```suggestion
stop_file = "/tmp/stopregservice"    # `touch` this file to stop the registration service
```

run/regservice.sh
```diff
@@ -0,0 +1,3 @@
+#!/bin/bash
```
oxinabox:
```suggestion
#!/bin/sh
```

run/restart.sh
```diff
@@ -0,0 +1,30 @@
+#!/bin/bash
```
oxinabox:
```suggestion
#!/bin/sh
```

run/commentbot.sh
christopher-dG: meta-suggestion: these should technically be `#!/usr/bin/env sh`

run/webui.sh
```diff
@@ -0,0 +1,3 @@
+#!/bin/bash
```
oxinabox:
```suggestion
#!/bin/sh
```

Dockerfile
```diff
@@ -11,7 +11,7 @@ USER registrator
 WORKDIR /home/registrator
 
 RUN julia -e "using Pkg; Pkg.add(PackageSpec(url=\"https://github.com/JuliaRegistries/Registrator.jl\"))"
```
christopher-dG:
```suggestion
RUN julia -e 'using Pkg; pkg"add Registrator#master; precompile"'
```

src/Messaging.jl
```diff
@@ -0,0 +1,144 @@
...
+
+"""
+Send an `::Any` object and wait for reply.
+
+Parameters:
+- `sock::RequestSocket`
+- `obj::Any`: The object to be sent
+
+Keyword Arguments:
+- `timeout::Real=60.0`: Timeout to wait for reply
```
oxinabox:
```suggestion
- `timeout::Real=60.0`: Time in seconds to wait for reply
```

src/Messaging.jl
```diff
@@ -0,0 +1,144 @@
...
+- `nretry::Number=Inf`: Number of retries after timeout
+
+Returns: Returns `nothing` if `nretry`s are up else returns received `::Any` object
+"""
+function sendrecv(sock::RequestSocket, obj; timeout::Real=60.0, nretry::Number=Inf)
+    send2zmq(sock.sock, obj)
+
+    while nretry != 0
```
oxinabox:
```suggestion
    while nretry > 0
```

README.web.md
```diff
@@ -101,6 +104,7 @@ It's important to note that optional values **must** be omitted or commented out
 - `disable_patch_notes`: Set to `true` to disable the patch notes text box.
 - `route_prefix`: Base route for the server.
   For example, use `/registrator` to serve the UI on `<your-hostname>/registrator/`.
+- `log_level`: The log level. Can be "INFO", "DEBUG", "WARN", "ERROR"
```
christopher-dG: This is apparently not optional, I got a `KeyError` when I ran without it.

src/Messaging.jl
```diff
@@ -0,0 +1,144 @@
...
+        rawfd = RawFD(sock.sock.fd)
+        istimedout = false
+        try
+            event = FileWatching.poll_fd(rawfd, timeout; readable=true)
+            istimedout = event.timedout
+        catch ex
+            # Socket was closed
+            ex isa EOFError && return nothing
```
oxinabox: Should we log something here?

run/config.web.toml
```diff
@@ -5,6 +5,10 @@ ip = ""
 port = 0
 server_url = ""
 registry_url = ""
+stop_file = "stopwebui"
+user = ""
+email = ""
+#log_level = "debug"
```
christopher-dG: a) see my earlier comment about maybe not being optional
b) lowercase "debug" is not one of the allowed values in the docs
c) `#foo` -> `# foo` please

src/Messaging.jl
```diff
@@ -0,0 +1,144 @@
...
+
+Returns:
+- `true`: If send was successful after receive
+- `false`: If socket was closed during receive
+"""
+function recvsend(f::Function, sock::ReplySocket)
+    # TODO: This should be done in a better way. Reasons why this was not done in a better way:
+    # 1) FileWatching.poll_fd does not poll reliable when waiting for a request.
+    # 2) zmq_poll hangs the julia process so no other tasks can run
+    while isopen(sock.sock) && sock.sock.events & ZMQ.POLLIN != ZMQ.POLLIN
```
oxinabox:
```suggestion
    while isopen(sock.sock) && sock.sock.events && ZMQ.POLLIN != ZMQ.POLLIN
```

src/RegService.jl
```diff
@@ -0,0 +1,41 @@
...
+using Logging
+
+import ..RegEdit: register
+using ..Messaging
+
+const config = Dict{String,Any}()
+
+include("management.jl")
+
+function service(zsock::ReplySocket)
```
oxinabox: Doc string here would be good

src/commentbot/CommentBot.jl
```diff
@@ -0,0 +1,252 @@
...
+    @info("Creating registration pull request for $(rp.reponame) PR: $(rp.trigger_src.prid)")
+end
+
+function print_entry_log(rp::RequestParams{CommitCommentTrigger})
+    @info("Creating registration pull request for $(rp.reponame) sha: `$(get_comment_commit_id(rp.evt))`")
+end
+
+function print_entry_log(rp::RequestParams{IssueTrigger})
+    @info("Creating registration pull request for $(rp.reponame) branch: `$(rp.trigger_src.branch)`")
+end
```
oxinabox: Should the above log messages use kwarg logging?
Similar ones below use kwarg logging.

Possible advantage is easier to computationally process.

src/commentbot/github_utils.jl
```diff
@@ -0,0 +1,249 @@
...
+            return payload["issue"]["html_url"]
+        end
+    elseif haskey(payload, "comment")
+        return payload["comment"]["html_url"]
+    else
+        error("Don't know how to get html_url")
+    end
+end
+
+function raise_issue(event, phrase, bt)
```
oxinabox: Docstring would bne nice here.
This does more things than I thought it would,
and the meaning of the  arg names are not clear to me

src/commentbot/param_types.jl
```diff
@@ -0,0 +1,172 @@
...
+        commenter_can_register = false
+        err = nothing
+        report_error = false
+
+        command = strip(phrase.captures[1], [' ', '`'])
+        action_name, action_args, action_kwargs = parse_comment(command)
+        target = get(action_kwargs, :target, nothing)
+
+        if evt.payload["repository"]["private"] && get(config["commentbot"], "disable_private_registrations", true)
+            err = "Private registration request received, ignoring"
```
oxinabox:
```suggestion
            err = "Private package registration request received, ignoring"
```

src/commentbot/parse_comment.jl
```diff
@@ -0,0 +1,26 @@
+const accept_regex = "([^\\r\\n]*)(\\n|\\r)*.*"
+
+function parse_comment(fncall)
+    argind = findfirst(isequal('('), fncall)
+    name = fncall[1:(argind - 1)]
+    parsed_args = Meta.parse(replace(fncall[argind:end], ";" => ","))
```
oxinabox: This smells like a security hole

src/management.jl
```diff
@@ -0,0 +1,79 @@
...
+            do_action()
+            backoff = 0
+        catch ex
+            exception_action = handle_exception(ex)
+            if exception_action == :exit
+                @warn("Stopping", name)
+                return
+            else # exception_action == :continue
+                bt = get_backtrace(ex)
+                @error("Recovering from unknown exception", name, ex, bt, backoff)
```
oxinabox: won't passing the `exception` kwarg to `@error` cause a backtrace to b e shown?
Or is that only in the `ConsoleLogger` ?

src/commentbot/parse_comment.jl
christopher-dG: This code long since in production, so this isn't really the place to comment on it (although it's hard to tell what is and isn't new with all the file renaming). Relevant: #130

src/regedit/register.jl
```diff
@@ -61,6 +61,12 @@ struct RegBranch
     end
 end
 
+function get_backtrace(ex)
+    v = IOBuffer()
+    Base.showerror(v, ex, catch_backtrace())
+    return v.data |> copy |> String
```
oxinabox: Rather than messing with `IOBuffers` you can use `sprint`
```
sprint(Base.showerror, ex, catch_backtrace())
````

File: src/management.jl [hunk 32]
```diff
@@ -0,0 +1,79 @@
...
+
+function status_monitor(stop_file::AbstractString,
+                        zsock::MessageSocket,
+                    )
+    while isopen(zsock.sock)
+        sleep(5)
+        flush(stdout); flush(stderr);
+        # stop server if stop is requested
+        if isfile(stop_file)
+            @warn("Server stop requested.")
```
oxinabox:
```suggestion
            @info("Server stop requested.")
```

src/management.jl
```diff
@@ -0,0 +1,79 @@
...
+function status_monitor(stop_file::AbstractString,
+                        event_queue::Channel,
+                        httpsock::Ref{Sockets.TCPServer},
+                    )
+    while isopen(event_queue)
+        sleep(5)
+        flush(stdout); flush(stderr);
+        # stop server if stop is requested
+        if isfile(stop_file)
+            @warn("Server stop requested.")
```
oxinabox:
```suggestion
            @info("Server stop requested.")
```

src/management.jl
```diff
@@ -0,0 +1,79 @@
...
+            flush(stdout); flush(stderr)
+
+            # stop accepting new requests
+            close(zsock.sock)
+
+            rm(stop_file; force=true)
+        end
+    end
+end
+
```
oxinabox: the two very similar functions above/below can be abstracted to use a common function that does the common stuff aboout handing `while cond; sleep(5); if stopfile ... rm`

src/management.jl
```diff
@@ -0,0 +1,79 @@
...
+function status_monitor(stop_file::AbstractString,
+                        zsock::MessageSocket,
+                    )
+    while isopen(zsock.sock)
+        sleep(5)
+        flush(stdout); flush(stderr);
+        # stop server if stop is requested
+        if isfile(stop_file)
+            @warn("Server stop requested.")
+            flush(stdout); flush(stderr)
```
oxinabox: Why are these being flushed manually?
That is rarely required.

src/webui/WebUI.jl
```diff
@@ -0,0 +1,227 @@
...
+    repo = @gf get_repo(forge, owner, name)
+    repo === nothing && error("Registry lookup failed")
+    clone = get(CONFIG, "registry_clone_url", url)
+    deps = get(CONFIG, "registry_deps", String[])
+    enable_patch_notes = !get(CONFIG, "disable_patch_notes", false)
+    REGISTRY[] = Registry(forge, repo, url, clone, deps, enable_patch_notes)
+end
+
+for f in [:index, :auth, :callback, :select, :register]
+    @eval $f(f::Function, r::HTTP.Request) = f($f(r))
```
oxinabox:
```suggestion
    @eval $f(func::Function, r::HTTP.Request) = func($f(r))
```

It is confusing to have two different `f`s

src/webui/WebUI.jl
```diff
@@ -0,0 +1,227 @@
...
+end
+
+error_handler(f::Function, r::HTTP.Request) = try
+    f(r)
+catch e
+    println(get_backtrace(e))
+    @error "Handler error" route=r.target
```
oxinabox: Why are we `println` inng that backtrace rather than using the logger?

src/webui/WebUI.jl
```diff
@@ -0,0 +1,227 @@
...
+    else
+        merge!(CONFIG, TOML.parsefile(config)["web"])
+    end
+
+    global_logger(SimpleLogger(stdout, get_log_level(CONFIG["log_level"])))
+    zsock = RequestSocket()
+
+    if haskey(CONFIG, "route_prefix")
+        for r in [:ROUTE_INDEX, :ROUTE_AUTH, :ROUTE_CALLBACK, :ROUTE_SELECT, :ROUTE_REGISTER, :ROUTE_STATUS]
+            @eval $r = CONFIG["route_prefix"] * $r
```
oxinabox: this is kinda gross. Is it required to be this way by MUX somehow?
If not, can we insead have a  global `const ROUTE = Dict(:INDEX=> ....
which is then mutated by this function.

src/webui/WebUI.jl
```diff
@@ -0,0 +1,227 @@
...
+        end
+    end
+
+    init_providers()
+    init_registry()
+
+    ip = CONFIG["ip"] == "localhost" ? Sockets.localhost : parse(IPAddr, CONFIG["ip"])
+    port = CONFIG["port"]
+
+    @info "Starting WebUI"
```
oxinabox:
```suggestion
    @info "Starting WebUI" ip port
```

src/webui/gitutils.jl
```diff
@@ -0,0 +1,159 @@
+# # Run some GitForge function, warning on error but still returning the value.
+macro gf(ex::Expr)
+    quote
+        let result = $(esc(ex))
+            GitForge.exception(result) === nothing ||
+                @warn "API request failed" exception=GitForge.exception(result)
+            GitForge.value(result)
```
oxinabox:
```suggestion
            return GitForge.value(result)
```

src/webui/gitutils.jl
```diff
@@ -0,0 +1,159 @@
...
+end
+
+# Get a repo's clone URL.
+cloneurl(r::GitHub.Repo) = r.clone_url
+cloneurl(r::GitLab.Project) = r.http_url_to_repo
+
+# Get a repo's tree hash.
+function gettreesha(f::GitHubAPI, r::GitHub.Repo, ref::AbstractString)
+    branch = @gf get_branch(f, r.owner.login, r.name, ref)
+    return branch === nothing ? nothing : branch.commit.commit.tree.sha
```
oxinabox: Is this correct that `commit.commit`?

src/webui/gitutils.jl
```diff
@@ -0,0 +1,159 @@
...
+        url = replace(url, host => "oauth2:$token@$host")
+    end
+
+    return try
+        mktempdir() do dir
+            dest = joinpath(dir, r.name)
+            run(`git clone $url $dest`)
+            match(r"tree (.*)", readchomp(`git -C $dest show $ref --format=raw`))[1]
+        end
+    catch
```
oxinabox: This is not great.
It catches *all* exceptions, and does not even log anything?
That means it could be catching `UndefinedVarError`s etc.

src/webui/gitutils.jl
christopher-dG: Yeah, it is.

oxinabox:
I have no idea how much of this is new content, since git is not great at tracking moved code.
(thus there is a recommendation to try really hard to not change code and move code in the same PR)

But seems like a good time to check over it anyway.

src/webui/gitutils.jl
christopher-dG: Yeah that is not ideal, this was basically a rough draft that I forgot to make nicer.

src/webui/gitutils.jl
christopher-dG: I do not think we want this. 
```julia
function foo()
    x = @gf get_user(gh)
    return x.login
```
This code will return `x`, the last line will never be reached.

oxinabox: ah, right, I forgot it was a macro.

src/Messaging.jl
nkottary: I meant to use bit wise AND here.

src/management.jl
nkottary: It does print the backtrace. But its all on one line. Difficult to read.

src/management.jl
nkottary: Required when using `nohup`

src/webui/WebUI.jl
nkottary: The logger shows newlines as `\n`.

src/management.jl
nkottary: `@warn` is appropriate here.

src/management.jl
nkottary: `@warn` is appropriate here.

src/management.jl
nkottary: Good point.

src/management.jl
oxinabox: comment as such?

src/webui/gitutils.jl
nkottary: I've edited this part to display the backtrace.

src/Messaging.jl
```diff
@@ -0,0 +1,144 @@
...
+        else
+            return recvfromzmq(sock.sock)
+        end
+        nretry -= 1
+    end
+    nothing
+end
+
+"""
+Receive an ::Any object from a socket, process it and send a reply.
```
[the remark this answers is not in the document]
nkottary: Fixed in c2691afc8ec69b7ebc6e9cca5bdc4eaada988d50

src/RegService.jl
nkottary: Fixed in c2691afc8ec69b7ebc6e9cca5bdc4eaada988d50

Don't know why it doesn't show as outdated.

src/commentbot/CommentBot.jl
nkottary: Fixed in f44f440fd04a4b6a458e33147a83e4bad92dfb41

Don't know why it doesn't show as outdated.

src/webui/routes/select.jl
nkottary: Why do we not want JavaScript? We use it here to ping the server for status of the registration. I don't understand how we can do this purely from the backend. The client will timeout if the clone or any other operation takes too long.

nkottary:
Concurrent registrations would be good to have but let's do that in a different PR once this is merged.

nkottary:
Will squash and rebase once the review is done.

src/webui/WebUI.jl
```diff
@@ -14,13 +14,13 @@ using TimeToLive
...
 using ..Messaging
 
-# Not const since they can be changed if $ROUTE_PREFIX is set.
-ROUTE_INDEX = "/"
-ROUTE_AUTH = "/auth"
-ROUTE_CALLBACK = "/callback"
-ROUTE_SELECT = "/select"
-ROUTE_REGISTER = "/register"
-ROUTE_STATUS = "/status"
+const ROUTE = Dict(:INDEX => "/",
```
christopher-dG: const _ROUTES_ perhaps?
Also format this way:
```
Dict(
    k => v,
)
```

src/webui/WebUI.jl
christopher-dG: This was displaying errors properly already... Does the different logger handle it differently?

src/webui/WebUI.jl
```diff
@@ -0,0 +1,221 @@
...
+            state = :errored
+        else
+            url = web_url(pr)
+            msg = """Registry PR successfully created, see it <a href="$url" target="_blank">here</a>!"""
+            state = :success
+        end
+    end
+    @debug msg
+    REGISTRATIONS[regdata.commit] = RegistrationState(msg, state)
+    nothing
```
christopher-dG: I don't think that explicitly returning `nothing` is really necessary

src/webui/WebUI.jl
```diff
@@ -0,0 +1,221 @@
...
+        error_handler,
+        pathmatch(ROUTE[:INDEX], index),
+        pathmatch(ROUTE[:AUTH], auth),
+        pathmatch(ROUTE[:CALLBACK], callback),
+        pathmatch(ROUTE[:SELECT], select),
+        pathmatch(ROUTE[:REGISTER], register),
+        pathmatch(ROUTE[:STATUS], status),
+        r -> html(404, "Page not found"),
+    )
+    do_action() = begin; t = serve(server, ip, port; server=httpsock[], readtimeout=0); wait(t); end
```
christopher-dG:
```suggestion
    do_action() = wait(serve(server, ip, port; server=httpsock[], readtimeout=0))
```

src/webui/gitutils.jl
```diff
@@ -0,0 +1,161 @@
...
+    end
+
+    return try
+        mktempdir() do dir
+            dest = joinpath(dir, r.name)
+            run(`git clone $url $dest`)
+            match(r"tree (.*)", readchomp(`git -C $dest show $ref --format=raw`))[1]
+        end
+    catch ex
+        println(get_backtrace(ex))
```
christopher-dG: Pretty sure this doesn't actually display the exception, only the backtrace.
Why not do the conventional `@error "This thing messed up" exception=(ex, catch_backtrace())` ? 
This is the same question that I had previously.

src/webui/routes/select.jl
christopher-dG: Plenty of people disable JS when possible, myself included. It's nice to avoid when possible. But we'll leave it as is and handle concurrency differently later.

src/webui/routes/select.jl
```diff
@@ -0,0 +1,87 @@
...
+        body *= """
+            Patch notes (optional):
+            <br>
+            <textarea cols="80" rows="10" name="notes"></textarea>
+            <br>
+            """
+    end
+
+    body *= """
+    <button id="submitButton" onclick="do_register()">Submit</button>
```
christopher-dG: This should be indented by 4 spaces

src/webui/gitutils.jl
nkottary: And I have already answered earlier that `Logging` does not print the stacktrace in a readable format. It prints it without the newlines. If you are using a better logger please let me know or add it to the Project file.

oxinabox: You can just change `SimpleLogger(io)` for `ConsoleLogger(io)` (from the Logging stdlib)
and you get Logging that looks just like the repl.

If color ends up being a problem (I can't recall if it naturally strips color if write to a stream,
then pass the kwarg:
```
meta_formatter = (args...) -> :normal
```
to the constructor

src/webui/WebUI.jl
```diff
@@ -14,13 +14,14 @@ using TimeToLive
 using Logging
 using ..Messaging
 
-const ROUTE = Dict(:INDEX => "/",
+const ROUTES = Dict(
```
christopher-dG: This is what I meant:
```julia
const ROUTES = Dict(
    :INDEX => "/",
    # ...
)
```

PR #139, opened 2019-05-07
Route prefix for Web UI

nkottary:


christopher-dG:
Is redefining consts not going to complain? What I would do would be to just set `ROUTE_PREFIX` to `get(ENV, "ROUTE_PREFIX", "")` and then do `const ROUTE_foo = "$ROUTE_PREFIX/foo"`.

christopher-dG:
Ok, I ran this locally and there are a few problems:

- The compiler yells at you about redefining constants
- The route prefix is decided at compile time and not runtime

See https://github.com/JuliaRegistries/Registrator.jl/commit/f28c9d0e4845a1f6258914fdd573c87ce7ca7ba0 for an implementation that seems to work.

nkottary:
Thank you for trying it out. f28c9d0 looks good. Can you push a PR with that?

Issue #137, opened 2019-05-07
Patch notes


christopher-dG:
Context: https://github.com/JuliaRegistries/TagBot/issues/8

So basically, the above issue proposes that you should be able to write your patch notes in your Registrator comment and have them propagated to your Git tag and GitHub release. I think this is a really nice idea.

Currently I don't see many people editing their TagBot-created GitHub releases with patch notes, and having the notes on both the tag and the release is a good practice. As is mentioned in the linked issue, Pkg could also start to display patch notes to users so that they know what to expect when upgrading stuff (I think I've seen package authors asking about ways to do this in the past as well).

Ideally, Registrator could copy the notes from the trigger point into its PR so that TagBot doesn't need to go looking for other commit comments or issues or PRs, not to mention the web UI doesn't have any point of reference to look back to. If not for this, the feature could be implemented solely on TagBot's end. 

There are a few syntax suggestions in the linked issue. I'm not really sure which one I like best, but #130 is somewhat relevant here because putting a multiline string in some pretend code would be a nightmare.

In any case, this is just to assess interest and if it seems like a good idea I can start implementing it on both Registrator and TagBot.

nkottary:
Fixed by #147

PR #141, opened 2019-05-07
Capture patch notes 

christopher-dG:
This is an implementation of #137, paired with https://github.com/JuliaRegistries/TagBot/pull/9. Right now it's just the web interface, since the GitHub bot is more complicated.

### DIFF
--- a/src/WebUI.jl
+++ b/src/WebUI.jl
@@ -33,7 +33,7 @@ const TEMPLATE = """
             margin: auto;
             max-width: 50em;
             font-family: Helvetica, sans-serif;
-            line-height: 1.5;
+            line-height: 1.8;
             color: #333;
           }
           a {
@@ -59,6 +59,10 @@ const PAGE_SELECT = """
     URL of package to register: <input type="text" size="50" name="package">
     <br>
     Branch to register: <input type="text" size="20" name="ref" value="master">
+    <br>
+    Patch notes (optional):
+    <br>
+    <textarea cols="80" rows="10" name="notes"></textarea>
     <br>
     <input type="submit" value="Submit">
     </form>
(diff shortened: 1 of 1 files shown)


src/Server.jl
```diff
@@ -737,8 +756,8 @@ function make_pull_request(pp::ProcessedParams, rp::RequestParams, rbrn::RegBran
     ref = get_html_url(rp.evt.payload)
 
     # FYI: TagBot (github.com/apps/julia-tagbot) depends on the "Repository", "Version",
-    # and "Commit" fields. If you're going to change the format here, please ping
-    # @christopher-dG and make sure that WebUI.jl has also been updated.
+    # "Commit", and "Patch notes" fields. If you're going to change the format here,
+    # please ping @christopher-dG and make sure that WebUI.jl has also been updated.
     params["body"] = """
```
nkottary: We should factor this out and make it a common method called by WebUI and the comment bot.

src/Server.jl
```diff
@@ -74,9 +76,14 @@ struct RequestParams{T<:RequestTrigger}
...
 
         if evt.payload["repository"]["private"] && get(config["registrator"], "disable_private_registrations", true)
-            err = "Private registration request recieved, ignoring"
+            err = "Private registration request received, ignoring"
             @debug(err)
         elseif action_name == "register"
+            # TODO:
+            # - The syntax with which users declare their patch notes is still undecided.
+            # - This assumes that patch notes appear last in the body. Is that fair?
+            patch_match = match(r"Patch notes:(.*)"s, body = get_body(evt.payload))
```
nkottary: You can't use kwarg `body = ` here. I think you meant to set `body` before calling match.

christopher-dG: Oops, yeah I had initially had `body = ...` in a separate line and  moved it in here. Thanks for catching.

PR #144, opened 2019-05-08
WebUI: Hide username and token if found in registry url

nkottary:


christopher-dG:
In what case would this be necessary?

christopher-dG:
If I'm guessing correctly,  I think setting `REGISTRY_CLONE_URL` might already deal with this.

PR #149, opened 2019-05-08
Fix GitLab permissions, method types, and remove unused method

### LINKED ISSUE #140: WebUI has no way to set depedency registries
Unlike the bot (thanks to #64 & #67), the WebUI has no way to use `registry_deps`, when it calls the `RegEdit.register` method.

The Bot server uses a config file, but the WebUI uses a series of environmental variables. Could we add a REG_DEPS environment variable?

I'm not sure how to pass array values as an ENV variable - when I've tried `export testvar=("hi", "testing")` and started julia, such a variable does not appear in ENV. Perhaps we could just use a string that is a comma-separated list of urls for the registry_deps?

I'm happy to make a PR implementing such a thing, and test the WebUI on Heroku if that is desirable.

christopher-dG:
First: Some GitLab permission checks were incorrect, I believe that they work properly now.

Second: I was passing in a `SubString` at some point which broke things, there's no downside to accepting other types of strings.

Third: The second change exposed some duplicate code which was probably the result of a bad rebase by me. I think I removed the right one, since the `RegistryCache` seems newer.

Closes #140

### DIFF
--- a/src/WebUI.jl
+++ b/src/WebUI.jl
@@ -170,10 +170,10 @@ end
 function isauthorized(u::User{GitLab.User}, repo::GitLab.Project)
     repo.visibility == "private" && return false
     forge = PROVIDERS["gitlab"].client
-    hasauth = @gf if repo.namespace == "user"
-        is_member(forge, repo.namespace.full_path, u.user.id)
-    else
+    hasauth = @gf if repo.namespace.kind == "user"
         is_collaborator(forge, repo.owner.username, repo.name, u.user.id)
+    else
+        is_member(forge, repo.namespace.full_path, u.user.id)
     end
     return something(hasauth, false)
 end
--- a/src/WebUI.jl
+++ b/src/WebUI.jl
@@ -1,6 +1,6 @@
 module WebUI
 
-using ..Registrator: RegEdit
+using ..Registrator: RegEdit, pull_request_contents
 
 using Base64
 using Dates
--- a/src/WebUI.jl
+++ b/src/WebUI.jl
@@ -21,27 +21,6 @@ ROUTE_REGISTER = "/register"
 
 const DOCS = "https://github.com/JuliaRegistries/Registrator.jl/blob/master/README.web.md#usage-for-package-maintainers"
 
-const PAGE_SELECT = """
-    <script>
-    function disableButton() {
-      var button = document.getElementById("submitButton");
-      button.disabled = true;
-      button.value = "Please wait...";
-    }
-    </script>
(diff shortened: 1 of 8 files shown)


src/WebUI.jl
```diff
@@ -170,10 +170,10 @@ end
 function isauthorized(u::User{GitLab.User}, repo::GitLab.Project)
     repo.visibility == "private" && return false
     forge = PROVIDERS["gitlab"].client
-    hasauth = @gf if repo.namespace == "user"
-        is_member(forge, repo.namespace.full_path, u.user.id)
-    else
+    hasauth = @gf if repo.namespace.kind == "user"
```
nkottary: If I was a collaborator but not a member of the org would this let me register?

src/regedit/register.jl
```diff
@@ -1,42 +1,8 @@
-"""
-Return a `GitRepo` object for an up-to-date copy of `registry`.
-"""
-function get_registry(registry::AbstractString; gitconfig::Dict=Dict())
```
nkottary: This is used here: https://github.com/JuliaRegistries/Registrator.jl/blob/03f4fa472ac6016e168271994110f9b1b53619ab/src/regedit/register.jl#L470

and here: https://github.com/JuliaRegistries/Registrator.jl/blob/03f4fa472ac6016e168271994110f9b1b53619ab/src/regedit/register.jl#L475

nkottary: Oh, I see that it is in types.jl. Nevermind.

src/WebUI.jl
christopher-dG: It would not, I can fix that.

christopher-dG:
I have also added a fix for #140, sorry to snipe you @BenJWard but I needed the fix for a deployed instance :sweat_smile:.

If we want to debate whether or not to use a config file over env variables, can we do it in a separate issue after getting this merged? I'd prefer to keep my deploy on master instead of a side branch as much as possible.

BenJWard:
> I have also added a fix for #140, sorry to snipe you @BenJWard but I needed the fix for a deployed instance 😅.
> 
> If we want to debate whether or not to use a config file over env variables, can we do it in a separate issue after getting this merged? I'd prefer to keep my deploy on master instead of a side branch as much as possible.

By all means, snipe away 😆 the main thing is the feature makes its way in!

christopher-dG:
This has gotten somewhat big -- sorry about that. I've basically been adding fixes here as issues come up on my own live deploy.

christopher-dG:
To please the big GitHub machine: Closes #140

nkottary:
What doesn't work? The registry deps option?

PR #152, opened 2019-05-09
WebUI: Use a TOML configuration file instead of environment variables

christopher-dG:
On top of #149, only commits including and after f98019571ffa024f10f46bdb5174fd25b1981884 are new.

This gives the web UI its own configuration file, without touching the existing GItHub bot's file.

Kind of WIP but as far as I can tell it's working identically to before.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,4 +1,4 @@
 /tmp
 /registries
 /packages
-*.env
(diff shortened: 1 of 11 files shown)


christopher-dG:
This would probably supersede #143

PR #155, opened 2019-05-13
Make patch notes capture case-insensitive

christopher-dG:
I'd expect plenty of people to use "Patch Notes:", etc.


StefanKarpinski:
Would "Release Notes" be a better trigger phrase? I feel like "Patch Notes" sounds like it only applies to patch releases, as compared to minor and major releases, whereas they are all releases of some kind.

christopher-dG:
We could just support both?

StefanKarpinski:
That seems fine to me. I would probably document and recommend one of them though.

PR #157, opened 2019-05-14
More user friendly error message, fix #156

### LINKED ISSUE #156: Changing package repo URL not allowed
Not able to register a release for ArrayFire because it has been moved from
https://github.com/JuliaComputing/ArrayFire.jl to https://github.com/JuliaGPU/ArrayFire.jl

"Error while trying to register: Changing package repo URL not allowed"

Old URL redirects to a new location, so it should be possible to detect and update package repo URL automatically.

gaika:


christopher-dG:
The target repo is not always JuliaRegistries/General, so the URL should be obtained from the config somewhere. I'm not quite sure what the right key is though.

src/regedit/register.jl
```diff
@@ -225,7 +225,7 @@ function find_package_in_registry(pkg::Pkg.Types.Project,
         package_path = joinpath(registry_path, package_data["path"])
         repo = TOML.parsefile(joinpath(package_path, "Package.toml"))["repo"]
         if repo != package_repo
-            err = "Changing package repo URL not allowed"
+            err = "Changing package repo URL not allowed, please submit a PR with the URL change to the registry at https://github.com/JuliaRegistries/General"
```
nkottary:
```suggestion
            u = HTTP.URI(registry)
            registry_url = u.scheme * "://" * u.host * u.path
            err = "Changing package repo URL not allowed, please submit a PR with the URL change to the registry at $registry_url"
```

nkottary: You will need to pass `registry` as an argument to `find_package_in_registry`

fredrikekre: Maybe we can exlude the specific URL in the message and avoid the extra function argument and other trickery. For example:
```suggestion
            err = "Changing package repo URL not allowed, please submit a pull request with the URL change to the target registry and retry."
```

PR #164, opened 2019-05-20
WebUI: Use Mustache.jl templates

nkottary:


src/webui/webutils.jl
```diff
@@ -13,43 +13,19 @@ function getcookie(r::HTTP.Request, key::AbstractString, default="")
     return ind === nothing ? default : cookies[ind].value
 end
 
+readtpl(tpl) = Mustache.template_from_file(joinpath(dirname(@__FILE__), "templates", tpl))
```
christopher-dG:
```suggestion
readtpl(tpl::AbstractString) = Mustache.template_from_file(joinpath(@__DIR__), "templates", tpl))
```

src/webui/webutils.jl
```diff
@@ -13,43 +13,19 @@ function getcookie(r::HTTP.Request, key::AbstractString, default="")
...
+const INDEX_TPL = readtpl("index.tpl")
+const SELECT_TPL = readtpl("select.tpl")
```
christopher-dG: I'm pretty sure making these things `const` means that they get compiled in and won't respond to updates in stuff like `route_prefix` etc.

src/webui/webutils.jl
```diff
@@ -13,43 +13,19 @@ function getcookie(r::HTTP.Request, key::AbstractString, default="")
...
-          <body>
-            <h1><a href="$(ROUTES[:INDEX])">Registrator</a></h1>
-            <h4>Registry URL: <a href="$registry" target="_blank">$registry</a></h3>
-            <h3>Click <a href="$DOCS" target="_blank">here</a> for usage instructions</h3>
-            <br>
-            $body
-          </body>
-        </html>
-        """
+    doc = render(INDEX_TPL,
```
christopher-dG: I would suggest:
```jl
render(
    INDEX_TPL;
    route_index=...,
)
```

src/webui/webutils.jl
christopher-dG: Also, are you sure you don't want `render_from_file`? I'm not positive myself but this is the exported function with a docstring.

src/webui/webutils.jl
nkottary: We pass routes as args to the templates while rendering so this is ok.

src/webui/webutils.jl
```diff
@@ -13,43 +13,21 @@ function getcookie(r::HTTP.Request, key::AbstractString, default="")
...
-          <body>
-            <h1><a href="$(ROUTES[:INDEX])">Registrator</a></h1>
-            <h4>Registry URL: <a href="$registry" target="_blank">$registry</a></h3>
-            <h3>Click <a href="$DOCS" target="_blank">here</a> for usage instructions</h3>
-            <br>
-            $body
-          </body>
-        </html>
-        """
+    doc = render_from_file(
```
christopher-dG: Hah, for future reference this is what I mean every time I bring it up:

```julia
doc = render_from_file(
    INDEX_TPL,  # Four spaces indented
    # ...
)  # Zero spaces indented
```

src/webui/routes/select.jl
```diff
@@ -1,88 +1,10 @@
...
-            <br>
-            """
-    end
-
-    body *= """
-        <button id="submitButton" onclick="do_register()">Submit</button>
-        </div>
-        """
-
+    body = render_from_file(
```
christopher-dG: same thing here -- don't worry about changing it now but just keep it in mind :slightly_smiling_face:

PR #170, opened 2019-05-27
README: use register() instead of register

visr:
Confused me in https://github.com/JuliaGeo/NetCDF.jl/commit/0f492e0c900b55fee3fac7183fc30cd0f31885aa

### DIFF
--- a/README.md
+++ b/README.md
@@ -16,14 +16,14 @@ Registrator is a GitHub app that automates creation of registration pull request
 First, install the app on your package(s) as mentioned above.  The procedure for registering a new package is the same as for releasing a new version.
 
 1. Set the [Project.toml](Project.toml) version field in your repository to your new desired `version`.
-2. Comment `@JuliaRegistrator register` on the commit/branch you want to register (e.g. like [here](https://github.com/JuliaRegistries/Registrator.jl/issues/61#issuecomment-483486641) or [here](https://github.com/chakravala/Grassmann.jl/commit/3c3a92610ebc8885619f561fe988b0d985852fce#commitcomment-33233149)).
+2. Comment `@JuliaRegistrator register()` on the commit/branch you want to register (e.g. like [here](https://github.com/JuliaRegistries/Registrator.jl/issues/61#issuecomment-483486641) or [here](https://github.com/chakravala/Grassmann.jl/commit/3c3a92610ebc8885619f561fe988b0d985852fce#commitcomment-33233149)).
 3. If something is incorrect, adjust, and redo step 2.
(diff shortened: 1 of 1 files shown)


StefanKarpinski:
This was intentionally changed in https://github.com/JuliaRegistries/Registrator.jl/pull/169 but the new version that supports the non-parens syntax may not be deployed yet. cc @nkottary

visr:
Ok, perhaps I was a bit too quick. Feel free to close, or merge then revert when it is live.

bgodard:
> This was intentionally changed in #169 but the new version that supports the non-parens syntax may not be deployed yet. cc @nkottary

Yes the non parens version is not working:
https://github.com/JuliaAstro/CALCEPH.jl/commit/6e61582c627830cf2387c3c794abd3846c750ab9#commitcomment-33690006

visr:
The current README with no parens is correct. This PR was reverted as soon as this new non-parens version went live. So if you try again now without parens it should work.

nkottary:
Both with and without parens is supported. Without parens is the old way. This is removed from documentation since we want to move to without parens as it is simpler.
 
Regarding https://github.com/JuliaAstro/CALCEPH.jl/commit/6e61582c627830cf2387c3c794abd3846c750ab9#commitcomment-33690006, Registrator is working. It has reported an error in the CALCEPH.jl Project file which needs to be addressed.

bgodard:
> Both with and without parens is supported. Without parens is the old way. This is removed from documentation since we want to move to without parens as it is simpler.
> 
> Regarding [JuliaAstro/CALCEPH.jl@6e61582#commitcomment-33690006](https://github.com/JuliaAstro/CALCEPH.jl/commit/6e61582c627830cf2387c3c794abd3846c750ab9#commitcomment-33690006), Registrator is working. It has reported an error in the CALCEPH.jl Project file which needs to be addressed.

That issue with JuliaRegistries was addressed immediately. I was referencing this comment thread because in the first comment JuliaRegistrator did not reply to my initial command without parens (before any edits) that I had copied from the  https://github.com/JuliaRegistries/Registrator.jl README.md.
So at least at that time two days ago that README file was not correct.

StefanKarpinski:
> So at least at that time two days ago that README file was not correct.

Yes, the timeline was:

1. The PR changing the syntax and the README was merged to master, but not deployed, so there was a brief period of time when the syntax in the README did not match what was actually supported. You encountered that brief period (no longer than six hours).

2. This PR was merged to bring the README back into agreement with what was deployed but the version on master still supported the new syntax, it just wasn't recommended in the README.

3. The master version was deployed with this PR reverted, so that the README and the deployed version both agree that the parensless syntax is both supported and preferred.

Unless there's any further confusion about this, I think we can stop talking about it now.

PR #172, opened 2019-05-28
Properly strip parens for multi-line input

christopher-dG:
Previously something like this:

```
register(branch="foo")
Some other text
```

Would parse `:branch => "foo)"`.

This is actually a non-issue for the current deploy because the input to the function never contains more than one line, but it's better to not make that assumption.

### DIFF
--- a/src/commentbot/parse_comment.jl
+++ b/src/commentbot/parse_comment.jl
@@ -17,7 +17,8 @@ The old syntax `action(key1=val1)` is supported for backwards compatibility.
 """
 function parse_comment(text::AbstractString)
     # Handling leading ( is easy, but not capturing the closing one is a bit harder.
-    text = join(map(line -> rstrip(line, [' ', ')']), split(text, "\n")), "\n")
+    lines = eachline(IOBuffer(text))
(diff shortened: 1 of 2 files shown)


src/commentbot/parse_comment.jl
```diff
@@ -17,7 +17,8 @@ The old syntax `action(key1=val1)` is supported for backwards compatibility.
 """
 function parse_comment(text::AbstractString)
     # Handling leading ( is easy, but not capturing the closing one is a bit harder.
-    text = strip(rstrip(text, ')'))
+    lines = eachline(IOBuffer(text))
+    text = join(map(line -> rstrip(line, [' ', ')']), lines), "\n")
```
omus: It is definitely not needed here but if this code does get to the point where bracket matching is required you can use recursive regular expressions to handle that case: `r"^(?<paren>\((?:[^()]*|(?&paren))\))$"`

omus: Could be simplified to:
```julia
    text = replace(text, r"[\)\s]+$"m => "")
```

PR #173, opened 2019-05-29
UI Enhancement

arunjose1995:
Few changes with ui was made.


christopher-dG:
The box should probably be made wider, and the centering seems off when I resize my browser to be half the width of my (16x9) screen.

2019-05-30-003328_421x342_scrot

StefanKarpinski:
We seem to have a chronic problem that all our UI people seem to use enormous screens and not test on small laptop screens 😁. I always have to complain a couple times that layout is broken for me.

christopher-dG:
Another screenshot for the curious 

2019-05-30-141954_1876x958_scrot

PR #180, opened 2019-06-14
Disable Codecov comments

christopher-dG:
Feel free to disagree, but I get no value from the comments it makes, and code coverage on this project seems to not be a priority in any case.

### DIFF
--- a/.codecov.yml
+++ b/.codecov.yml
@@ -1,0 +1,1 @@
+comment: off


christopher-dG:
welp, guess it doesn't work...

fredrikekre:
Yea, I have never gotten this to work...

christopher-dG:
maybe this works...

christopher-dG:
I've been doing some ad hoc testing and I'm not positive, but I think Codecov only reads config from the master branch. So if we merge this then _maybe_ it'll stop yelling at us in future PRs.

PR #195, opened 2019-07-04
Include part of the UUID in branch names

### LINKED ISSUE #165: Branch name Registrator creates can have a name collision
The branch registrator creates is of the form `package_name/version`. However, there might be multiple package with the same name. So should probably include a part of the UUID in the branch name as well.

fredrikekre:
Use `Base.version_slug` for branch names, fixes #165, fixes https://github.com/JuliaRegistries/General/issues/1765.

### DIFF
--- a/src/regedit/utils.jl
+++ b/src/regedit/utils.jl
@@ -14,4 +14,6 @@ end
 
 Generate the name for the registry branch used to register the package version.
 """
-registration_branch(pkg::Pkg.Types.Project) = "register/$(pkg.name)/v$(pkg.version)"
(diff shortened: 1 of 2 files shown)


fredrikekre:
Maybe it should be
```
registrator/$(lowercase(pkg.name))/$(slug)
```
?

fredrikekre:
Actually, we should probably not use the tree-sha for this; then there would be new PRs all the time. Should compute something based on uuid + version instead.

fredrikekre:
Just included the 8 first characters of the UUID instead.

Issue #194, opened 2019-07-03
refuse to register already-tagged versions


StefanKarpinski:
There are a lot of package authors still tagging first and registering later, which often leads to mismatches between what’s tagged and what’s registered. TagBot posts a comment after the fact warning that this has happened, but we should check if a tag or release already exists and refuse to kick off registration in the first place if it doesn’t point to the commit that is being registered.

nkottary:
In the README we mention that the registration process does not depend on the tag. I would say that the package owners are responsible for the tag being correct.

StefanKarpinski:
The registration process doesn't depend on the tag, but we do need tags in order to be sure that the registered tree stays around even if the commit is no longer on any public branch, so we do kind of depend on tags. It's also just bad if there are tags/releases and they don't match what's registered. It's fine if there's no tag (as long as there's a public branch that doesn't go away), but if there is a tag and it's wrong, that's actively confusing.

nkottary:
Okay.

nkottary:
The issue is fixed but not deployed yet.

nkottary:
Should we allow existing tags if the tag commit matches the registration commit?

StefanKarpinski:
Yes, I think it's fine in that case.

DilumAluthge:
> Should we allow existing tags if the tag commit matches the registration commit?

What happens if the initial registration matches the existing tag, but then a General registry maintainer requests a change (e.g. "please add a compat entry for Julia and then re-trigger Registrator")? After the package maintainer makes the requested change and re-triggers Registrator, the updated registration will no longer match the existing tag.

nkottary:
We are checking the commit id of the tag in #203 . When they don't match the registration will not go through.

StefanKarpinski:
> What happens if the initial registration matches the existing tag, but then a General registry maintainer requests a change (e.g. "please add a compat entry for Julia and then re-trigger Registrator")? After the package maintainer makes the requested change and re-triggers Registrator, the updated registration will no longer match the existing tag.

Then they have to delete the existing tag.

stev47:
> Then they have to delete the existing tag.

In general (non-julia) one creates a patch-release to fix the previous broken release. From the point of Registrator.jl this will look like a skipped version number since it didn't accept the broken one. Would it be feasible to allow the version skip in that case?

adigitoleo:
[...]

PR #202, opened 2019-07-10
WebUI: Check for existing tag

nkottary:
Address #194 for WebUI

### DIFF
--- a/src/webui/gitutils.jl
+++ b/src/webui/gitutils.jl
@@ -226,3 +226,44 @@ mention(u::GitLab.User) = "@$(u.username)"
 # If the registry is from the same provider, mention the user, otherwise use the URL.
 display_user(u::U) where U =
     (parentmodule(typeof(REGISTRY[].repo)) === parentmodule(U) ? mention : web_url)(u)
+
+function tagexists(forge::GitHubAPI, repo::GitHub.Repo, tag::VersionNumber)
+    result = @gf get_tags(forge, repo.owner.login, repo.name)
+
+    if result === nothing
+        @debug("Could not fetch tags")
+        return false
+    end
+
+    for t in result
+        v = split(t.ref, "/")[end]
+        if startswith(v, "v")
+            v = v[2:end]
+        end
+        v == string(tag) && return true
+    end
+    false
+end
+
+function tagexists(forge::GitLabAPI, project::GitLab.Project, tag::VersionNumber)
+    # Using Registrator's token here instead of the users. This is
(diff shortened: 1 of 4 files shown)


src/webui/gitutils.jl
```diff
@@ -226,3 +226,44 @@ mention(u::GitLab.User) = "@$(u.username)"
...
+        if startswith(v, "v")
+            v = v[2:end]
+        end
+        v == string(tag) && return true
+    end
+    false
+end
+
+function tagexists(forge::GitLabAPI, project::GitLab.Project, tag::VersionNumber)
+    # Using Registrator's token here instead of the users. This is
```
christopher-dG: FYI pretty much every non-write GitLab interaction is done with the Registrator API key instead of the user's, so this comment isn't really necessary in this particular spot. I thought I had it written down in a comment somewhere but perhaps not.

src/webui/gitutils.jl
```diff
@@ -226,3 +226,44 @@ mention(u::GitLab.User) = "@$(u.username)"
...
+    end
+
+    for t in result
+        v = split(t.ref, "/")[end]
+        if startswith(v, "v")
+            v = v[2:end]
+        end
+        v == string(tag) && return true
+    end
+    false
```
christopher-dG:
```suggestion
    return false
```

src/webui/gitutils.jl
```diff
@@ -226,3 +226,44 @@ mention(u::GitLab.User) = "@$(u.username)"
...
+    end
+
+    for t in result
+        v = t.name
+        if startswith(v, "v")
+            v = v[2:end]
+        end
+        v == string(tag) && return true
+    end
+    false
```
christopher-dG:
```suggestion
    return false
```

src/webui/gitutils.jl
nkottary: I left it as a note to myself when I have to support registration for private packages.

src/webui/gitutils.jl
nkottary: I typically use `return` to return and break out of a block. If its the last line I avoid using an extra keyword.

christopher-dG: Sure, makes sense.

nkottary:
I don't know why the test is failing: `ERROR: empty intersection between GitForge@0.1.3 and project compatibility 0.1.4-*`

fredrikekre:
You probably need to update the manifest.

Issue #74, opened 2019-04-09
How to auto-login with GIT?


BenJWard:
Following fixing #71, my Heroku installation of Registrator, now gets further in the process, however, it halts as at some point GIT asks for a username and password:

```
2019-04-09T14:38:08.021638+00:00 app[web.1]: ┌ Debug: commit changes

2019-04-09T14:38:08.021644+00:00 app[web.1]: └ @ Registrator /home/registrator/.julia/packages/Registrator/DcyDi/src/register.jl:357

2019-04-09T14:38:08.031292+00:00 app[web.1]: ┌ Debug: push -f branch to remote

2019-04-09T14:38:08.031295+00:00 app[web.1]: └ @ Registrator /home/registrator/.julia/packages/Registrator/DcyDi/src/register.jl:370

2019-04-09T14:38:08.293122+00:00 app[web.1]: fatal: could not read Username for 'https://github.com': No such device or address
```

Is there something I have to do to get it to log in with the GitHub user automatically?

From reading around I can think of a few ways of doing it, one is to use ssh instead of http.

Another is to try and use the API token in the config to push to `git push https://4UTHT0KEN@github.com/foo/bar` instead of `git -C /home/registrator/registries/ccbd2cc2-2954-11e9-1ccf-f3e7900901ca -c user.name=BioJuliaRegistrar -c user.email=benwardbiologist@gmail.com push -q -f -u origin register/FASTX/v0.1.0`

@nkottary What do you think is the best way to get around this? I'm happy to submit a PR with any needed fixes.

BenJWard:
So I've played with this a bit more and I've found that if I set the `repo` argument in the conf.toml to the form of `https://USER:TOKEN@github.com/User/Repo`, then the server will successfully complete a registration request when running on Heroku. I don't know if this is the best way of doing it, or instead, using a deploy key assigned specifically to the registry repo.

nkottary:
I think `https://USER:TOKEN@github.com/User/Repo` is fine. Without that `git` is going to ask for password when pushing to the registry branch.

BenJWard:
The user token does work. Would you consider it the server supporting using a Deploy Key for the registry repo? Documenter uses the deploy key method, and unlike an API token, it limits the push permission to the single repository.

On the other hand, the API token used by the user is already needed for commenting and making PRs so it might as well be used to clone.

[...]

PR #208, opened 2019-07-23
ENV Variable configuration of comment bot for Heroku

BenJWard:
This PR adds the ability to get the port from ENV for the WebUI.
It also addresses  #74, but allowing URLs for registries to be specified through ENV.

This is useful for services like Heroku where the URL used for the repository contains the access token: `https://USER:TOKEN@github.com/User/Repo`.

I'm not sure, but currently, I don't think the token field in `config.commentbot.toml` is actually used.
Because providing an access token for this field, and then using a normal https URL for the registries results in authentication errors for me, whereas using repo urls of the form, `https://USER:TOKEN@github.com/User/Repo` does not.

### DIFF
--- a/src/commentbot/CommentBot.jl
+++ b/src/commentbot/CommentBot.jl
@@ -132,8 +132,10 @@ function action(rp::RequestParams{T}, zsock::RequestSocket) where T <: RegisterT
     
     # Here we will check if the target registry has a repo argument, and if not,
     # assign it a default one, obtained from ENV.
-    if !haskey(target_registry, "repo") && haskey(ENV, "DEFAULT_REGISTRY_URL")
-        target_registry["repo"] = ENV["DEFAULT_REGISTRY_URL"]
+    if !haskey(target_registry, "repo")
+        specific_env_key = string(uppercase(target_registry_name), "_REGISTRY_URL")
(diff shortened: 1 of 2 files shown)


BenJWard:
There is an alternative to this, which is to do something like this in the Docker:

`RUN git config --global url.”https://{token}:@github.com/".insteadOf “https://github.com/"`

If you'd rather that I try that and see if that works before changing Registrator code, that's cool, I can try that this week.

src/commentbot/CommentBot.jl
```diff
@@ -129,6 +129,14 @@ function action(rp::RequestParams{T}, zsock::RequestSocket) where T <: RegisterT
             target_registry_name, target_registry = filteredtargets[1]
         end
     end
+    
+    # Here we will check if the target registry has a repo argument, and if not,
+    # assign it a default one, obtained from ENV.
+    if !haskey(target_registry, "repo")
+        specific_env_key = string(uppercase(target_registry_name), "_REGISTRY_URL")
+        envkey = ifelse(haskey(ENV, specific_env_key), specific_env_key, "DEFAULT_REGISTRY_URL")
```
nkottary: This and the next line can be simplified to:

```
target_registry["repo"] = get(ENV, specific_env_key, ENV["DEFAULT_REGISTRY_URL"])
```

PR #212, opened 2019-08-22
Allow `Registrator` to push directly to `master`

staticfloat:


staticfloat:
Not that this is going to be used for the actual Registrator bot, but when testing locally (and pushing to your own fork) this is really useful.

src/regedit/register.jl
```diff
@@ -504,8 +504,10 @@ function register(
         @debug("branch registry repo")
         git = gitcmd(registry_path, gitconfig)
         run(pipeline(`$git checkout -f master`; stdout=devnull))
-        run(pipeline(`$git branch -f $branch`; stdout=devnull))
-        run(pipeline(`$git checkout -f $branch`; stdout=devnull))
+        if branch != "master"
+            run(pipeline(`$git branch -f $branch`; stdout=devnull))
```
christopher-dG: Unrelated to this PR, but we could use `checkout -fb` instead of `branch` and `checkout` in separate commands.

PR #215, opened 2019-09-06
Copy over https://github.com/JuliaLang/Pkg.jl/pull/1350

staticfloat:


src/regedit/register.jl
```diff
@@ -160,18 +160,23 @@ subset of the pool of available versions, this function computes a `VersionSpec`
...
     for b in reverse(subset)
         a.major == b.major || continue
-        for m = 1:3
-            lo = VersionBound((a.major, a.minor, a.patch)[1:m]...)
-            for n = 1:3
-                hi = VersionBound((b.major, b.minor, b.patch)[1:n]...)
+        for m = 1:4
+            lo = VersionBound((a.major, a.minor, a.patch, a.build)[1:m]...)
+            for n = 1:4
+                hi = VersionBound((b.major, b.minor, b.patch, b.build)[1:n]...)
```
davidanthoff: This is a bug: if `VersionNumber` doesn't have `build` info, then `x.build` is an empty tuple, but there is no  method for `VersionBound` that accepts empty tuples. I have a situation right now where that crashes things. @staticfloat could we either revert this PR in the short run to unblock the situation I'm trying to solve, or come up with a fix?

PR #224, opened 2019-09-27
Revert "Copy over https://github.com/JuliaLang/Pkg.jl/pull/1350"

davidanthoff:
This reverts #215, which introduced a bug (see my comment on that PR) that is throwing when I use my rewrite of PkgDev.jl (but it looks like a generic bug to me that should affect others as well).

A speedy merge would be fantastic.

CC @staticfloat

### DIFF
--- a/src/regedit/register.jl
+++ b/src/regedit/register.jl
@@ -160,12 +160,7 @@ subset of the pool of available versions, this function computes a `VersionSpec`
 includes all versions in `subset` and none of the versions in its complement.
 """
 function compress_versions(pool::Vector{VersionNumber}, subset::Vector{VersionNumber})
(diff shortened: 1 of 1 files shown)


staticfloat:
Sorry @davidanthoff, I've been pretty buried under BB stuff recently.  I think the "proper" fix here is to indeed drop the second hunk, but to keep the first one.  E.g. the diff I should have applied is simply this one: https://github.com/JuliaLang/Pkg.jl/pull/1350/files so the `VersionBounds()` constructors should never have been touched.

davidanthoff:
Cool! I added those, so this PR now just reverts the parts that touch the `VersionBounds` constructor calls. I think this is probably the proper fix, right?

staticfloat:
Yep, looks good to me!

PR #228, opened 2019-10-02
Add julia compat

fredrikekre:
I vaguely recall that Registrator requires Julia 1.1, can someone confirm? @nkottary maybe?


ararslan:
FWIW my colleagues and I are running it internally with 1.2. I think it's just 1.0 (and probably nightly, based on CI) that doesn't work.

fredrikekre:
Alright. This is probably correct then.

PR #235, opened 2019-10-16
Update links to documentation in readme

notZaki:
The readme has some links to the documentation but some of them were giving me a 404 error. 
I've tried to update those cases---I used the stable version of the documentation and made a guess as to which page the links should point to.

Also, the very last sentence of the readme was removed since it appeared to be redundant.


fredrikekre:
Thanks.

PR #266, opened 2020-02-07
Make trigger phrase case-insensitive

christopher-dG:
This lets us use `@juliaregistrator register` (fixes a gripe mentioned on Slack).

### DIFF
--- a/src/commentbot/CommentBot.jl
+++ b/src/commentbot/CommentBot.jl
@@ -203,10 +203,12 @@ function comment_handler(event::WebhookEvent, phrase::RegexMatch)
     return HTTP.Messages.Response(200)
 end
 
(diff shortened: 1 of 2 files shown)


nickrobinson251:
Could do with a test?

christopher-dG:
I'm not actually sure how to approach that. The trigger is passed into GitHub.jl's `CommentListener` and not something that Registrator owns.

DilumAluthge:
Maybe a simple test that confirms that this regex does match `JuliaRegistrator` and `juliaregistrator`, and does not match `ZuliaRegistrator` or `zuliaregistrator`.

christopher-dG:
Good idea!

christopher-dG:
I'm gonna go ahead and merge this tomorrow if no one has objections.

PR #280, opened 2020-04-20
Fix raise_issue

SebRollen:
`raise_issue` currently takes a `Regex` as its second argument, but the code itself expects a `RegexMatch`, as the code later tries to find the `match` field of `phrase`. This PR simply modifies the method signature so that the method can be matched

### DIFF
--- a/src/commentbot/github_utils.jl
+++ b/src/commentbot/github_utils.jl
@@ -102,7 +102,7 @@ function get_html_url(payload::Dict{<:AbstractString})
 end
 
 """
-    raise_issue(event::WebhookEvent, phrase::Regex, bt::String)
+    raise_issue(event::WebhookEvent, phrase::RegexMatch, bt::String)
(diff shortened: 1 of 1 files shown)


SebRollen:
Just to check that there are no places where we'd expect to see `raise_issue` called with a `Regex` param, I went through to find where `raise_issue` is invoked, which is currently in three places:
1) https://github.com/JuliaRegistries/Registrator.jl/blob/b790c217767069ca3dd4127f0a122ced5269b6cf/src/commentbot/CommentBot.jl#L172
2) https://github.com/JuliaRegistries/Registrator.jl/blob/b790c217767069ca3dd4127f0a122ced5269b6cf/src/commentbot/CommentBot.jl#L200
3) https://github.com/JuliaRegistries/Registrator.jl/blob/838cbd91f6cf7c341373401fab420f02230003b7/src/commentbot/approval.jl#L134

In all cases, the second argument seems to be a `RegexMatch`

PR #281, opened 2020-04-20
Update stop-* make commands to correct file paths

SebRollen:
Currently, the default config for the commentbot, regservice and webui is to listen for file-creations in the /tmp/ folder, but the default behavior for the make command is to create the stop-files in the working directory. This pr changes the latter to create the files in /tmp/, as expected

### DIFF
--- a/run/config.commentbot.toml
+++ b/run/config.commentbot.toml
@@ -8,7 +8,7 @@ backend_port = 5555             # Port of the backend regservice
 trigger = "@JuliaRegistrator"   # The comment that should trigger Registrator. Typically
                                 # keep this the same as your GitHub user mention.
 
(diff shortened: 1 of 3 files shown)


nkottary:
I was thinking about this and I felt the best thing to do here would be to change the default path for the stop file in the config file to the current working directory.

The reason is that this will let you run multiple Registrators without having to edit the Makefile and the stop file paths in the config file(s). Running multiple Registrators is useful when you update Registrator.jl package and you want to make sure that everything works by bringing up a staging Registrator.

SebRollen:
That's a good point. I updated the PR to instead create and listen for the stopfiles in the working directory

PR #282, opened 2020-04-20
Negate some comments

SebRollen:
The inline comments next to some of the options here are confusing as they currently describe the opposite of the behavior specified by the config file.

### DIFF
--- a/run/config.commentbot.toml
+++ b/run/config.commentbot.toml
@@ -23,16 +23,16 @@ set_status = true               # If set to true Registrator will post pending a
 enc_key = "1234567890123456"    # The encryption key to use to encrypt PR metadata. This is
                                 # decrypted and used by Registrator for the approved() call.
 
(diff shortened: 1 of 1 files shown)


nkottary:
Can you change `enable` to `disable` instead of `true` to `false` in the comment? Since the default is `false` I would say that it would read better this way.

SebRollen:
Sure! I updated the comments to reflect that.

It still reads kind of strangely as the comments say that enabling the pull request trigger and approval call is not recommended, yet it's the default option. Perhaps a better fix is to change the actual option in the config file to default to `disable_pull_request_trigger = true` and `disable_approval_process = true`?

nkottary:
Yes, I agree.

SebRollen:
Okay, I updated the defaults to be `true`, and updated the comments to reflect what happens if they're set to `false`. The behavior / language seems much more clear now, in my opinion

DilumAluthge:
@nkottary Is this good to merge now?

PR #293, opened 2020-05-22
Added support for ssh addresses.

### LINKED ISSUE #289: Failing to register new version of package using ssh address
Context: In our company we have private registry, and self-hosted registrator, and we use ssh keys for authentication for our self-hosted github.
Problem: I can not register new version of package when Registry is already using ssh address using the web UI.
In our registry, for packages we have e.g.
`Package.toml` in the registry
```toml
name = "InternalPackage"
uuid = "a2c437b0-bf6b-41a1-8a62-162a814afa99"
repo = "git@git.our_company.com:our_org/InternalPackage.jl.git"
```
when I put `git@git.our_company.com:our_org/InternalPackage.jl.git` to the form https://github.com/JuliaRegistries/Registrator.jl/blob/master/src/webui/templates/select.tpl
I obtain: ERROR: Package URL is invalid.

when I put `https://git.our_company.com:our_org/InternalPackage` to the form https://github.com/JuliaRegistries/Registrator.jl/blob/master/src/webui/templates/select.tpl
I obtain: ERROR: Registration failed: Changing package repo URL not allowed, please submit a pull request with the URL change to the target registry and retry.

Since we don't have anonymous clone allowed, we are using SSH keys everywhere and thus having https address would complicate things for us a lot.

racinmat:
Fixes #289

### DIFF
--- a/src/webui/routes/register.jl
+++ b/src/webui/routes/register.jl
@@ -12,8 +12,16 @@ function register(r::HTTP.Request)
     form = parseform(String(r.body))
     package = get(form, "package", "")
     isempty(package) && return json(400; error="Package URL was not provided")
-    occursin("://", package) || (package = "https://$package")
-    match(r"https?://.*\..*/.*/.*", package) === nothing && return json(400; error="Package URL is invalid")
+    is_ssh = startswith(package, "git@") && occursin(":", package)
+    if !is_ssh && !occursin("://", package)
+         package = "https://$package"
+    end
+    if endswith(package, ".git")
+        package = package[1:end-length(".git")]
+    end
+    !is_ssh && match(r"https?://.*\..*/.*/.*", package) === nothing && return json(400; error="Package URL is invalid")
+    is_ssh && match(r"git@.*\..*:.*/.*", package) === nothing && return json(400; error="Package URL is invalid")
+
     ref = get(form, "ref", "")
     isempty(ref) && return json(400; error="Branch was not provided")
     notes = get(form, "notes", "")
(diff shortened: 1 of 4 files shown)


racinmat:
I see codecov tests are not passing. But since the whole library is kinda weirdly tested, should I do something with it, or should I let it be?

src/webui/routes/register.jl
```diff
@@ -12,8 +12,16 @@ function register(r::HTTP.Request)
...
     package = get(form, "package", "")
     isempty(package) && return json(400; error="Package URL was not provided")
-    occursin("://", package) || (package = "https://$package")
-    match(r"https?://.*\..*/.*/.*", package) === nothing && return json(400; error="Package URL is invalid")
+    is_ssh = startswith(package, "git@") && occursin(":", package)
+    if !is_ssh && !occursin("://", package)
+         package = "https://$package"
+    end
+    if endswith(package, ".git")
+        package = package[1:end-length(".git")]
```
nkottary, line 20: Why was this necessary? Did you get an error when you tried registering or do you not use .git in your company?

If this is necessary then it needs to be handled here https://github.com/JuliaRegistries/RegistryTools.jl/blob/03096523fe864f86ee374019a2ebea461e47d9be/src/register.jl#L331 as well (i.e do the equality check without ".git") or else it will give an error for registrations done using the publicly deployed registrator.

racinmat: It was necessary because the GitHub API called from forge returned 404 to me when it contained .git.

nkottary: I just tested out this PR and it seems to be fine. Sorry for the noise.

PR #310, opened 2020-11-15
Add the `check_authorization` kwarg to the `Registrator.WebUI.register` function

DilumAluthge:
This pull request makes the following changes:
1. Adds the `check_authorization::Bool` keyword argument to the `Registrator.WebUI.register` function. The default value is `check_authorization = true`.
2. Bumps the version number in `Project.toml` from `1.2.2` to `1.3.0`.

## Motivation

I am working on an alternate way of deploying Registrator.jl that does not require a persistent server to be running. As part of this, I hook into Registrator at the `register` function. However, I check that the user is authorized to make releases of the package before I call this function, so there is no need to repeat this check inside the `register` function. Therefore, I need to be able to disable this check.

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -1,7 +1,7 @@
 name = "Registrator"
 uuid = "4418983a-e44d-11e8-3aec-9789530b3b3e"
 authors = ["Stefan Karpinski <stefan@karpinski.org>"]
-version = "1.2.2"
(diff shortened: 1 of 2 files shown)


DilumAluthge:
I may have found a different way to accomplish what I need.

Issue #329, opened 2021-04-14
Permissions for Registrator


aviks:
So we've had unhappy users complain that the web registrator hosted on juliahub.com asks for write permissions. I wanted to discuss the options, since this is security stance that affects the registry itself. We need consensus from the registry maintainers about what the correct stance should be. 

The underlying requirement is that before the registrator user opens a PR on the registry repo, it needs to be sure that the user requesting the registration has write permissions to the repository. 

For the web registrator we need to query github apis to get this information. However, the list of "collaborators" on a package is not available to a user unless that user has write permissions to the repository. This is the reason we ask for write permissions for the Oauth token we get from Github. Also, afaik, we cannot restrict permissions per repository on the Oauth token -- per repository access can be given using a deploy token, but that is not a one-click flow. 

As a result of the strength of said feedback, we have this morning changed juliahub web registrator to use only read permissions. I fear however, this may have been a knee jerk reaction by us. To enable the registrator functionality with read permissions, we are now using the list of "contributors" for a repo.  This information is freely accessible. However, this list also includes any users who have had a PR merged to the repo, even if they do not have write access. Additionally, this list precludes any organisation members or admins, who may have write permissions, but have not actually merged any code to that repo. Both of which is problematic. The latter can be solved by asking such users to make their org membership public, but the former (user with PR merged) is a problem. 

The commentbot needs the same information, but it works as a Github app that is installed on each repo on which it operates. As a result, it has the permissions required to get these priviledged fields. 

So I suppose the main question to the registry maintainers is as follows: are we happy to have a security stance where all "contributors" in a repo can make a registry PR. 

Given that most registry PRs are merged automatically, and with the focus on "supply chain attacks", this is an important question that should be carefully considered.

fredrikekre:
> web registrator hosted on juliahub.com asks for write permissions.

This is nothing new though, right?

> are we happy to have a security stance where all "contributors" in a repo can make a registry PR

I don't think so. Wouldn't that essentially mean "everyone" (i.e. you just have to fix a type in the docs or soemthing) can make releases?

DilumAluthge:
> I don't think so. Wouldn't that essentially mean "everyone" (i.e. you just have to fix a type in the docs or soemthing) can make releases?

Agree. This seems like a big security hole.

aviks:
> This is nothing new though, right?

[...]

PR #330, opened 2021-04-15
WIP: Option to authorize registrations from a file in the package repo

nkottary:
This lays down the ground work for the plan that is still under discussion in https://github.com/JuliaRegistries/Registrator.jl/issues/329. This does not change the existing authorization behaviour of the WebUI so existing installations will not be affected. The file based authorization is enabled by a configuration parameter which is optional.

I've written down the details in the dev docs. We need to update the README.md after this goes live on juliahub.com since users refer to that for help on using Registrator.


StefanKarpinski:
One comment, let's not call this `authorised_registrars.txt`, which imo is a bit of an odd name. If this file is going to exist in a lot of repos it seems good to match existing conventions for similar files — how about `MAINTAINERS` to match the `CODEOWNERS` file name convention? We should also figure out the file format.

PR #334, opened 2021-05-03
Fix getting repo namespace for GitLab

nkottary:
`repo.namespace` is always present but `repo.owner` can sometimes be `nothing`


DilumAluthge:
I'm not as familiar with the GitLab API, but it looks right to me.

BTW, why is CI not running?

fredrikekre:
Probably the free Travis time is used already

PR #347, opened 2021-08-11
passing subdirectory to RegistryTools

### LINKED ISSUE #346: Registration of subdirectory using web UI does not create subdir field in Package.toml
When using the Web UI of registrator after the https://github.com/JuliaRegistries/Registrator.jl/pull/344 fix, the resulting Package.toml does not contain the `subdir` field. I think it's because in https://github.com/JuliaRegistries/Registrator.jl/blob/master/src/webui/WebUI.jl#L132 there is not `subdir` parameter, but in https://github.com/JuliaRegistries/Registrator.jl/blob/master/src/commentbot/CommentBot.jl#L147 there is the parameter. 

Should I send a PR for it?

racinmat:
Fixes #346

### DIFF
--- a/src/webui/WebUI.jl
+++ b/src/webui/WebUI.jl
@@ -58,6 +58,7 @@ struct RegistrationData
     commit::String
     notes::String
     is_ssh::Bool
+    subdir::String
 end
 
 const event_queue = Channel{RegistrationData}(1024)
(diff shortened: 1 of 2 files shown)


christopher-dG (approved):
LGTM, but I'll let someone else pull the trigger.

nkottary (approved):

nkottary:
Thank you!

PR #349, opened 2021-08-31
add description field

ericphanson:
Over in https://github.com/JuliaRegistries/General/issues/43776 it was suggested that it might be nice to have the repo description field in the PR comment. We could then grab it from the PR comment and send it downstream as well, e.g. to the Slack channel `#new-packages-feed`, along with the release notes.

From the WebUI or CommentBot we seem to have a `repo` object with this field, so we can get the description without an additional API call. However, I'm not sure if it is always populated or not; if it is not populated, then we'd need to make an API call to fill out the fields (`repo` in the comment bot to use GitHub.jl, or `get_repo` in the WebUI to use GitForge.jl).

### DIFF
--- a/src/commentbot/CommentBot.jl
+++ b/src/commentbot/CommentBot.jl
@@ -75,6 +75,8 @@ function make_pull_request(pp::ProcessedParams, rp::RequestParams, rbrn::RegBran
                   "maintainer_can_modify"=>true)
     ref = get_html_url(rp.evt.payload)
 
+    description = something(rp.evt.repository.description, "")
+
     params["title"], params["body"] = pull_request_contents(;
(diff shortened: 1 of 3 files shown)


DilumAluthge:
Who is a good person to review this?

@nkottary?

DilumAluthge (approved):
This looks good to me, but it would be good to get a review from someone that is more familiar with the Registrator codebase.

nkottary (approved):

nkottary:
The new version of Registrator v1.2.9 that has this change is not yet deployed. I am very busy this week. I will try to get this updated next week.

nkottary:
Just an update: We're working to get this to production. Might take a few more days.

PR #360, opened 2022-01-19
Add release notes note if no release notes provided

IanButterworth:
Untested, but should look something like this:
____

```
@JuliaRegistrator register()
```
____

Registration pull request created: JuliaRegistries/General/xxxx

### Tip: Release Notes
        
Did you know you can add release notes too? Just add markdown formatted text underneath the comment and it will be added to the registry PR, and if tagbot is installed it will also be added to the tag created on this repository.

### Tagging
After the above pull request is merged, it is recommended that a tag is created on this repository for the registered package version.

This will be done automatically if the Julia TagBot GitHub Action is installed, or can be done manually through the github interface, or via:
```
git tag -a v1.3.1 -m "<description of version>" 7d4c8c4c918148afc371e4814d1f2929c3f46b37
git push origin v1.3.1
```

### DIFF
--- a/src/commentbot/CommentBot.jl
+++ b/src/commentbot/CommentBot.jl
@@ -97,25 +97,37 @@ function make_pull_request(pp::ProcessedParams, rp::RequestParams, rbrn::RegBran
 
     pr, msg = create_or_find_pull_request(repo, params, rbrn)
     tag = tag_name(ver, subdir)
-    
-    releasenotes_note if isempty(rp.release_notes)
+
+    releasenotes_note = if isempty(rp.release_notes)
         """
-        
+
         ### Tip: Release Notes
-        
-        Did you know you can add release notes too? Just add markdown formatted text underneath the comment and it will be added to 
-        the registry PR, and if tagbot is installed it will also be added to the tag created on this repository.
-        
+
+        Did you know you can add release notes too? Just add markdown formatted text underneath the comment with the header
+        "Release notes:" and it will be detected and added to the registry PR, and if tagbot is installed it will also
+        be added to the tag created on this repository.
+
+        i.e.
+        ```
+        @JuliaRegistrator register()
+
+        Release notes:
+
+        ## Breaking changes
+
+        - Foo
+        ```
+
         """
     else
         ""
     end
-    
+
     cbody = """
         Registration pull request $msg: [$(repo)/$(pr.number)]($(pr.html_url))
         $(releasenotes_note)
         ### Tagging
-    
+
         After the above pull request is merged, it is recommended that a tag is created on this repository for the registered package version.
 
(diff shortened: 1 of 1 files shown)


src/commentbot/CommentBot.jl
```diff
@@ -94,10 +94,25 @@ function make_pull_request(pp::ProcessedParams, rp::RequestParams, rbrn::RegBran
...
     pr, msg = create_or_find_pull_request(repo, params, rbrn)
     tag = tag_name(ver, subdir)
-
+    
+    releasenotes_note if isempty(rp.release_notes)
+        """
+        
+        ### Tip: Release Notes
+        
+        Did you know you can add release notes too? Just add markdown formatted text underneath the comment and it will be added to
```
ericphanson, line 103: I think this isn't enough; you need to format it to pass this regex: https://github.com/JuliaRegistries/Registrator.jl/blob/1558fb7a627a85eee6b58950e507daf54efb9dfe/src/commentbot/param_types.jl#L49

like described here: https://github.com/JuliaRegistries/Registrator.jl#release-notes.

But maybe we could relax that regex, or just mention they should write `"Release notes:"`

IanButterworth: Ah ok. I updated it. If JuliaRegistrator writes `@JuliaRe...` as an example, will that cause the universe to collapse?

DilumAluthge: If you put the ping inside of backticks, JuliaRegistrator shouldn't actually get pinged, so you should be good there.

DilumAluthge: But if the ping isn't in backticks, I actually do think that the bot will get caught in a loop replying to itself 😂 

And eventually hit its rate limit, causing all registrations to grind to a halt.

ericphanson: uhhh maybe 🤔. I think in https://github.com/JuliaRegistries/Registrator.jl/blob/1558fb7a627a85eee6b58950e507daf54efb9dfe/src/commentbot/CommentBot.jl#L188 we could bail if the user is JuliaRegistrator. Or maybe here: https://github.com/JuliaRegistries/Registrator.jl/blob/1558fb7a627a85eee6b58950e507daf54efb9dfe/src/commentbot/param_types.jl#L20. Not totally sure though

DilumAluthge: Yeah I was just thinking that. I'll try to make a PR.

DilumAluthge: https://github.com/JuliaRegistries/Registrator.jl/pull/361

IanButterworth:
Now that TagBot is fixed can we add this? 
See also https://discourse.julialang.org/t/make-breaking-change-releases-easier-to-handle-add-release-notes-to-tags-via-juliaregistrator/103241

ericphanson (approved):

PR #361, opened 2022-01-20
Make sure that the `JuliaRegistrator` bot never tries to at-ping itself

DilumAluthge:


nkottary:
@DilumAluthge Can you also call your `mention` method on `evt.sender` right after this [line](https://github.com/JuliaRegistries/Registrator.jl/blob/9d589154624f59d421706dc44f32704b401b0836/src/commentbot/CommentBot.jl#L194)? That will ensure that we throw an error if Registrator calls itself.

DilumAluthge:
> Can you also call your `mention` method on `event.sender` right after this [line](https://github.com/JuliaRegistries/Registrator.jl/blob/9d589154624f59d421706dc44f32704b401b0836/src/commentbot/CommentBot.jl#L194)? That will ensure that we throw an error if Registrator calls itself.

Done! I put it right after the `@debug` statement, since I figure that debug message might be useful sometimes.

ericphanson:
I think instead of erroring if it messages itself, it should just ignore it. That way we can mention ourselves in #360. So instead of adding a `mention` function I think we could just early-exit from the comment handler if it’s self-triggered.

IanButterworth:
Maybe exit even before starting julia? I believe the user could be checked in a job and skipped there?

PR #362, opened 2022-01-20
Support SSH urls for the registry

nkottary:
This lets you use an SSH url for your registry repo when you configure the WebUI server.


vdayanand (approved):
LGTM

PR #364, opened 2022-01-21
RFC: CLI entrypoint

arnaudh:
I'm creating this PR to get some feedback on the idea of having a Command Line Interface (CLI) entrypoint to Registrator.

The use case is that we (at [Invenia](https://github.com/invenia)) want to be able to automatically trigger the registration of package versions when a PR gets merged to the main branch of any repo we have. This would be for our internal GitLab packages, and we'd be using GitLab CI's features to trigger this.

Currently the Registrator has two "entrypoints": the WebUI and the GitHub app. We've been using the WebUI (self-hosted), and while it has been working well, it does add a manual step whenever we release a new package version, which happens often with our ~100 internal packages. Adding automation on top of the WebUI seemed complex and unnecessary, hence why I thought it could make sense to have a simple CLI, which could be used with something like:

```
julia CLI.jl --config path/to/config.toml --package https://gitlab.company.com/Example.jl/ --git-ref master --notes "..."
```

We'd be able to use this in our CI jobs, triggered automatically when PRs are merged. Hopefully this would also benefit others who want to automate registration.

My questions are:

1. Could there be issues in the approach?
  I'm wondering for example if there could be concurrency issues, if we were to trigger the CLI.jl script multiple times in parallel (on different processes or machines). I'm wondering because I see message queues are used internally, and I'm wondering if they serve a concurrence safety purpose (e.g. is it important that only one git process is active at once, or only one registry PR is created at once?). The CLI.jl I propose doesn't use those queues, it's one single-threaded running script.
  
2. Does it make sense to add to Registrator.jl?
  To me it does because it mainly just re-uses the logic found in other files. In the current PR, I've mostly just copied over the code (with a comment indicating where I copied it from), except for one function which I've extracted to see what it would look like to avoid code duplication: `Registrator.WebUI.build_registration_data` (should be moved outside of WebUI).
 
Overall code is very rough and untested, just to show the idea. If we're happy with the approach, I'll submit a proper PR when code is polished and tested.
  
I very appreciate any feedback on this!

### DIFF
--- a/src/CLI.jl
+++ b/src/CLI.jl
@@ -1,0 +1,116 @@
+
+using GitForge, GitForge.GitHub, GitForge.GitLab
+using Logging
+using RegistryTools
+using TOML
+include("Registrator.jl")
+
+using Registrator.WebUI: get_log_level
+
+function main(config::AbstractString=isempty(ARGS) ? "config.toml" : first(ARGS))
+    merge!(Registrator.WebUI.CONFIG, TOML.parsefile(config)["web"])
+    if get(Registrator.WebUI.CONFIG, "enable_logging", true)
+        global_logger(SimpleLogger(stdout, get_log_level(get(Registrator.WebUI.CONFIG, "log_level", "INFO"))))
+    end
+
+    @show Registrator.WebUI.CONFIG
+
+    # Based on Registrator.WebUI.init_providers()
+    gitlab = Registrator.WebUI.CONFIG["gitlab"]
+    Registrator.WebUI.PROVIDERS["gitlab"] = Registrator.WebUI.Provider(;
+        name="GitLab",
+        client=GitLabAPI(;
+            url=get(gitlab, "api_url", GitLab.DEFAULT_URL),
+            token=PersonalAccessToken(gitlab["token"]),
+            has_rate_limits=!get(gitlab, "disable_rate_limits", false),
+        ),
+        client_id=gitlab["client_id"],
+        client_secret=gitlab["client_secret"],
+        auth_url=get(gitlab, "auth_url", "https://gitlab.com/oauth/authorize"),
+        token_url=get(gitlab, "token_url", "https://gitlab.com/oauth/token"),
+        scope="read_user api",
+        include_state=false,
+        token_type=OAuth2Token,
+    )
+
+    Registrator.WebUI.init_registry()
+    
+    register()
+end
+
+
+function register()
+    pkey = "gitlab"
+
+    provider = get(Registrator.WebUI.PROVIDERS, pkey, nothing)
+    # client = typeof(provider.client)(;
+    #     url=GitForge.base_url(provider.client),
+    #     token=provider.token_type(token),
+    #     has_rate_limits=GitForge.has_rate_limits(provider.client, identity),
+    # )
+    client = provider.client # Not sure what I'm doing here. Skipping oauth token dance I guess?
+    u = Registrator.WebUI.User(@Registrator.WebUI.gf(get_user(client)), client)
+    @show client
+    @show u
+
+    # TODO read these from CLI args
+    package = "https://gitlab.company.com/Example.jl/"
+    ref = "main"
+    notes = ""
+    subdir = ""
+    regdata = Registrator.WebUI.build_registration_data(u, package, ref, notes, subdir)
+
+
+    # Based on WebUI.jl action()
+    regp = RegistryTools.RegisterParams(
+        Registrator.WebUI.cloneurl(regdata.repo, regdata.is_ssh), 
+        regdata.project, 
+        regdata.tree;
+        subdir=regdata.subdir,
+        registry=Registrator.WebUI.REGISTRY[].clone, 
+        registry_deps=Registrator.WebUI.REGISTRY[].deps, 
+        push=true,
+    )
+
+    # Based on RegService.jl service()
+    # TODO read from config
+    regp.gitconfig["user.name"] = "Registrar"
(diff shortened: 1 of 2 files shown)


fredrikekre:
Are you aware of https://github.com/GunnarFarneback/LocalRegistry.jl ?

arnaudh:
> Are you aware of https://github.com/GunnarFarneback/LocalRegistry.jl ?

I wasn't aware of this, thank you for the link. While it looks simpler and close to our use case, it doesn't seem to support our workflow which is to submit a PR to our internal registry (like Registrator does) and run CI tests on it (using [RegistryCI.jl](https://github.com/JuliaRegistries/RegistryCI.jl)) before the PR is merged.

GunnarFarneback:
Cf. https://github.com/GunnarFarneback/LocalRegistry.jl/issues/45

All functionality that's missing is to make the actual PR. Basically
```
# `path` is local path to the package. In a CI setting probably the current directory.
# `registry` is a URL to the registry, possibly including write credentials.
# `branch` is the branch name you want to use for the PR.
if LocalRegistry.do_register(path, registry; branch)
    # Create the PR.
end
```

GunnarFarneback:
Out of curiosity,

> `julia CLI.jl --config path/to/config.toml`

Is the idea that this config file contains registry url, gitlab user id, credentials, etc? Would there be a copy of this file in every package repository or do you envision some mechanism to centralize it?

GunnarFarneback:
For reference my current approach at work requires this for the package CI when the conditions for registration are met:
```
julia -e 'using Pkg; Pkg.add("XXXXXX")'
julia --project -e 'using XXXXXX; register(user = "${BITBUCKET_CI_USER}", token = "${BITBUCKET_CI_TOKEN}")
```
where `XXXXXX` is a private package which uses `LocalRegistry` to perform the registration, makes REST calls to our Bitbucket server to create a registry PR, and hardcodes the registry URL.

I'm not super happy about the invocation but at least it's somewhat short and there are no duplicated config files. The obvious drawback is that it requires a custom package, which doesn't generalize very easily. I'm looking for a way to combine `LocalRegistry` and `GitForge` for a more general solution.

arnaudh:
> Out of curiosity,
> 
> > `julia CLI.jl --config path/to/config.toml`
> 
> Is the idea that this config file contains registry url, gitlab user id, credentials, etc? Would there be a copy of this file in every package repository or do you envision some mechanism to centralize it?

This config is the one needed to run the web service for Registrator.jl (see [config.web.toml template](https://github.com/arnaudh/Registrator.jl/blob/master/run/config.web.toml)), indeed it contains registry url, gitlab user id, credentials. There is only one across the org and we have it stored centrally in AWS S3.

> And is CLI.jl the file from this Registrator PR? 

Yes.

> How is the package CI supposed to get hold of it?

The way this is setup / being setup:
1. All repos in our private GitLab have a CI pipeline (defined in a `.gitlab-ci.yml` file) that extends a common template. This template (among other things) defines on the default branch (master/main) a trigger to the pipeline below, passing appropriate `REPO_URL` and `GIT_REFERENCE`:
2. There is an empty `registrator` repo that just defines a CI pipeline that takes as input `REPO_URL` and `GIT_REFERENCE` env variables. The pipeline runs a single job which:
    1. Clones [Registrator.jl](https://github.com/JuliaRegistries/Registrator.jl/) (currently using this PR's branch)
    2. Downloads the config from S3
    3. Runs
    ```
    julia --project=. -e 'using Pkg; Pkg.instantiate()'
    julia --project=. src/CLI.jl --config ./config.toml --package "$REPO_URL" --git-ref "$GIT_REFERENCE"
    ```
    This creates a PR to our internal registry
3. The PR gets approved & marked for auto-merge by a self-hosted [TagBotGitLab](https://github.com/invenia/TagBotGitLab)
4. Once the pipeline passes (runs [RegistryCI.jl](https://github.com/JuliaRegistries/RegistryCI.jl]) on the PR), it gets merged 
5. Once merged, TagBotGitLab will additionally create a GitLab release on the package
  
Effectively this gives us full release automation, from the moment a new version MR is merged to the default branch of any given Julia repo.

arnaudh:
I am not opposed to using LocalRegistry.jl instead of Registrator.jl (for step 2 above), however that will require quite some extra work, to implement things in LocalRegistry, and to adapt to it.

Implementing the PR creation in LocalRegistry.jl means:
1. deciding which to support: GitLab/GitHub/Bitbucket, and whether to use GitForge.jl (sounds like the way to go, though doesn't support Bitbucket as of now)
2. implement: configuration, fetching of repo information (to create PR body), creation of PR

If we do this, there will be quite a bit of duplication with Registrator.jl, so I'm wondering if there is a higher level discussion to be had, to re-combine these packages in some way (e.g. create/extract a third common package).

GunnarFarneback:
Work in progress with git push options in https://github.com/GunnarFarneback/LocalRegistry.jl/pull/49. There are some complications but the idea is that for GitLab a package CI should only need to do
```
julia -e 'using Pkg; Pkg.add("LocalRegistry")'
julia --project -e 'using LocalRegistry; register(registry = $REGISTRY_URL, create_gitlab_mr = true)'
```
to get a merge request which will be automatically merged once the registry CI passes.

arnaudh:
Coming back to this a bit late, but this is to say we've been using LocalRegistry.jl since @GunnarFarneback has implemented the GitLab PR creation logic mentioned above. The workflow is working perfectly for us: packages are being registered automatically and it reduced our development friction by quite a bit. 

Massive thanks for implementing this @GunnarFarneback! 🙏

PR #365, opened 2022-01-25
Adding Bitbucket support

zot:
This relies on new Bitbucket support in GitForge: https://github.com/JuliaWeb/GitForge.jl/pull/35


nkottary:
Done in #392

PR #379, opened 2022-07-14
Fix constructor

nkottary:


DilumAluthge (approved):
Maybe add a note in the commit message or PR title that we are adding a missing `subdir` argument?

PR #383, opened 2022-08-19
Added section on deal with common error messages

eford:


nkottary (approved):

DilumAluthge (changes requested):
I don't think that this is quite the right way to go. It doesn't make sense to duplicate content from the General registry README in this README. Instead, we should link directly to the General registry README. That way, we only need to update the content in one place.

PR #392, opened 2022-09-29
Support Bitbucket

zot:
This PR adds support for Bitbucket, as well as some source documentation and refactoring needed for Bitbucket support -- Bitbucket diverges from many of the similarities between GitHub and GitLab.

### DIFF
--- a/src/webui/WebUI.jl
+++ b/src/webui/WebUI.jl
@@ -144,7 +144,7 @@ function init_registry()
     	# not using @gf here so we can log the error
         repo, _ = get_repo(forge, owner, name)
     catch err
-        error("Registry lookup failed$(err === nothing ? "" : err)")
+        error("Registry lookup failed: $err")
     end
     clone = get(CONFIG, "registry_clone_url", url)
     fork_url = get(CONFIG, "registry_fork_url", clone)
--- a/src/webui/WebUI.jl
+++ b/src/webui/WebUI.jl
@@ -4,7 +4,8 @@ using ..Registrator: pull_request_contents
 import RegistryTools
 
 using Dates
-using GitForge, GitForge.GitHub, GitForge.GitLab
+using GitForge, GitForge.GitHub, GitForge.GitLab, GitForge.Bitbucket
+using Base64
 using HTTP
 using JSON
 using Mux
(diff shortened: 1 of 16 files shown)


zot:
Hmm, tests are failing but not locally -- I'll get that working...

src/webui/WebUI.jl
```diff
@@ -92,6 +115,11 @@ include("routes/callback.jl")
 include("routes/select.jl")
 include("routes/register.jl")
 include("routes/status.jl")
+include("routes/bitbucket.jl")
+
+auth_href(::Provider{GitLabAPI}) = "$(ROUTES[:AUTH])?provider=gitlab"
+auth_href(::Provider{GitHubAPI}) = "$(ROUTES[:AUTH])?provider=github"
+auth_href(::Provider{BitbucketAPI}) = "$(ROUTES[:BITBUCKET])/auth"
```
nkottary, line 122: Can't we use the same format `/auth/?provider=bitbucket`?

src/webui/WebUI.jl
```diff
@@ -102,16 +130,22 @@ function init_registry()
...
     owner, name = splitrepo(url)
-    repo = @gf get_repo(forge, owner, name)
-    repo === nothing && error("Registry lookup failed")
-
+    repo = nothing
+    try
+    	# not using @gf here so we can log the error
+        repo, _ = get_repo(forge, owner, name)
+    catch err
+        error("Registry lookup failed$(err === nothing ? "" : err)")
```
nkottary, line 147: We don't need the `nothing` check here. `err` will never be `nothing`.

src/webui/WebUI.jl
zot: No, Bitbucket puts its own query parameters in the URL -- it was probably a bad idea in the first place to assume it was safe to use query parameters

src/webui/WebUI.jl
zot: fixed in the latest commit

src/webui/WebUI.jl
zot: I did try that in the beginning, btw -- if I remember correctly, bitbucket just replaced our parameters with its own

src/webui/gitutils.jl
```diff
@@ -40,130 +66,180 @@ is_success(res::AuthFailure) = false
...
+    ((@gf_bool @mock is_collaborator(u.forge, repo.owner.login, repo.name, u.user.login)) ||
+        (@gf_bool @mock is_collaborator(jforge, repo.owner.login, repo.name, u.user.login))) &&
+        return AuthSuccess()
+    # If the repo is not in an org, the user does not have sufficient access to create a PR
+    repo.organization === nothing &&
+        return AuthFailure("User $(u.user.login) is not a collaborator on repo $(repo.name)")
+    # Otherwise the user must be a member of the organization
+    # verify with user's connection, falling back to JuliaHub's connection
+    (something((@gf_q @mock is_member(u.forge, repo.organization.login, u.user.login)), false) ||
+        something((@gf_q @mock is_member(jforge, repo.organization.login, u.user.login)), false)) &&
```
nkottary, line 92: These two lines can be simplified. `something((@gf_q f()), false)` is same as `@gf_bool f()`.

src/webui/gitutils.jl
```diff
@@ -40,130 +66,180 @@ is_success(res::AuthFailure) = false
...
-    forge = PROVIDERS["gitlab"].client
-    result = nothing
+    forge = provider(repo).client
+    lasterr = nothing
+    for file in Base.project_names
+        try
+            fc, _ = get_file_contents(forge, repo.id, joinpath(subdir, file); ref=ref)
+            return decodeb64(fc.content)
+        catch err
+            lasterr = err
```
nkottary, line 172:
```suggestion
            lasterr = (err, catch_backtrace())
```

src/webui/gitutils.jl
zot: pushed that change

nkottary (approved):

PR #400, opened 2022-12-23
Fix some regressions

nkottary:
- UndefVarError for `mention` in comment bot
- lowercased provider in callback.jl so that provider can be found in config file

### DIFF
--- a/src/commentbot/github_utils.jl
+++ b/src/commentbot/github_utils.jl
@@ -281,3 +281,5 @@ function create_or_find_pull_request(
 
     return pr, msg
 end
+
+mention(u::GitHub.Owner) = "$(mention(u.login))"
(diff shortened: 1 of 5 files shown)


src/commentbot/github_utils.jl
```diff
@@ -281,3 +281,5 @@ function create_or_find_pull_request(
 
     return pr, msg
 end
+
+mention(u::GitHub.Owner) = "$(mention(u.login))"
```
zot, line 285: should there be a `@` in there?

```suggestion
mention(u::GitHub.Owner) = "@$(mention(u.login))"
```

zot: oh wait -- it's another call to mention...

nkottary:
See #401

PR #402, opened 2022-12-23
Fix kc/pkgpro

nkottary:


KristofferC:
I believe https://github.com/JuliaRegistries/Registrator.jl/pull/397/commits/502cb22d42f4c5769692bda4cf799199e5ba00d0 is the correct fix.

PR #406, opened 2023-01-18
Change assertion to "if" and log error

nkottary:
The assertion that is removed in this PR has failed recently for a user. The changes in this PR should give more information if such a failure happens again


zot (approved):
Looks good!

PR #407, opened 2023-01-20
Add note for private packages

nkottary:
- Change `@gf_q` to `@gf` for a `get_repo` call so that error is logged
- Add docs for allowing third party OAuth access to private repositories on GitHub


zot (approved):
Looks good to me!

PR #408, opened 2023-02-06
Fix WebUI tag check for subdir packages

nkottary:
~The `istagwrong` method checks whether the commit SHA of the registration ref matches the commit SHA of the corresponding tag (if such a tag exists). This does not work for subdir packages since their versions need not be same as the repo tag. This PR skips the tag check when subdir is non-empty~


nkottary:
I see that the commentbot searches for tags with the sub directory as prefix. So instead of disabling this check I think its better to fix webui to also search for tags with the prefix. I will update this PR.

PR #428, opened 2023-12-03
Don't use the `exception=get_backtrace(...)` pattern

mortenpi:
It makes the stacktraces from the error handlers pretty much useless and hinders debugging. In particular, since `exception` is now just a string, the logging macros will elide most of the stacktrace. As an MWE:

```
julia> try
       error("1234")
       catch e
       @error "Error handler" exception = sprint(Base.showerror, e, catch_backtrace())
       end
┌ Error: Error handler
│   exception = "1234\nStacktrace:\n  [1] error(s::String)\n    @ Base ./error.jl:35\n  [2] top-level scope\n    @ REPL[9]:2\n  [3] eval\n    @ ./boot.jl:370 [inlined]\n  [4] eval_user_input(ast::Any, backend::REPL.REPLBackend, mod::Module)\n    @ REPL ~/.julia/juliaup/julia-1.9" ⋯ 974 bytes ⋯ "816 [inlined]\n [12] run_main_repl(interactive::Bool, quiet::Bool, banner::Bool, history_file::Bool, color_set::Bool)\n    @ Base ./client.jl:405\n [13] exec_options(opts::Base.JLOptions)\n    @ Base ./client.jl:322\n [14] _start()\n    @ Base ./client.jl:522"
└ @ Main REPL[9]:4
```

Instead I suggest using the standard `exception = (e, catch_backtrace())` pattern.

### DIFF
--- a/src/commentbot/CommentBot.jl
+++ b/src/commentbot/CommentBot.jl
@@ -210,8 +210,8 @@ function action(rp::RequestParams{T}, zsock::RequestSocket) where T <: RegisterT
             set_error_status(rp)
         end
     catch ex
-        bt = get_backtrace(ex)
-        @info("Unexpected error: $bt")
+        @info "Unexpected error" exception = (ex, catch_backtrace())
+        bt = sprint(Base.showerror, ex, catch_backtrace())
         raise_issue(rp.evt, rp.phrase, bt)
     end
     @info("Done processing register event", reponame=rp.reponame, target_registry_name)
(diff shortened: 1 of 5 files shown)


src/commentbot/CommentBot.jl
```diff
@@ -210,8 +210,8 @@ function action(rp::RequestParams{T}, zsock::RequestSocket) where T <: RegisterT
             set_error_status(rp)
         end
     catch ex
-        bt = get_backtrace(ex)
-        @info("Unexpected error: $bt")
+        @info "Unexpected error" exception = (ex, catch_backtrace())
+        bt = sprint(Base.showerror, ex, catch_backtrace())
```
mortenpi, line 214: This introduces a tiny bit of duplication in two places. But I feel that if `raise_issue` should be doing some special printing, then there should probably be a special method for that function.

nkottary (approved):

PR #448, opened 2025-06-13
safer `withpasswd` method

tanmaykm:
The `withpasswd` method executes a shell script with dynamic variables interpolated. The values could sometimes contain characters that may break script syntax. Those could, depending on how an application is using the APIs, also be coming from user inputs, so they can be unpredictable. This change is to interpolate the base64 values of the variables which would be safe and then pipe it to base64 decode utility at runtime to print out the unencoded string. Also added a test.

This brings in additional dependency on the `base64` utility though, which is probably not too bad.


nkottary (approved):

DilumAluthge (approved):
LGTM

tanmaykm:
Yes please! I do not have permissions to merge to this repo.

PR #451, opened 2025-06-16
add more tests for cloneurl valid pattern

tanmaykm:
Add more path traversal tests.
Ref: https://github.com/JuliaRegistries/Registrator.jl/pull/449#discussion_r2148838047


DilumAluthge:
Thank you @tanmaykm and @nkottary!

PR #455, opened 2025-06-20
`Manifest.toml`: Upgrade to `HTTP@1.10.17`, `URIs@1.6.0`, and `OpenSSL@1.5.0`

DilumAluthge:
I used Julia 1.7.2 when updating the manifest.

This PR was created with:

```
] add --preserve=all HTTP@1.10.17 URIs@1.6.0 OpenSSL@1.5.0

] rm OpenSSL
```

Note: We need to upgrade OpenSSL to at least `>= 1.3`, otherwise we can't upgrade to `HTTP@1.10.17`.

### DIFF
--- a/Manifest.toml
+++ b/Manifest.toml
@@ -44,6 +44,12 @@ version = "3.46.0"
 deps = ["Artifacts", "Libdl"]
 uuid = "e66e0078-7015-5450-92f7-15fbd957f2ae"
 
+[[deps.ConcurrentUtilities]]
+deps = ["Serialization", "Sockets"]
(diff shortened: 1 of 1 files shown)


DilumAluthge:
cc: @nkottary @tanmaykm

tanmaykm (commented):
Do we need to keep the Manifest.toml file in the package repo? Would it be better to remove it altogether?

DilumAluthge:
> Do we need to keep the Manifest.toml file in the package repo? Would it be better to remove it altogether?

Yep, I think we should answer that question first: https://github.com/JuliaRegistries/Registrator.jl/issues/456

tanmaykm:
^ @nkottary would probably know?

PR #480, opened 2026-04-01
Blocklist functionality

### LINKED ISSUE #443: add mechanism to block users
in General we have seen some spam AI generated registration requests from the same user. It would be great if Registrator could block users. General itself can't block users since the PRs come via registrator, so blocking a user from the repo doesn't block their registration requests.

This could work by making Registrator check a private github repo (with a configurable URL for other deployments) that has a list of users or something like that.

rayegun:
Heavily vibe-coded by Claude Opus 4.6, but examined line by line manually.

Intended to resolve #443 by adding a simple configurable blocklist check to both the commentbot and the webui. The blocklist is a (private) repo with a TOML file, and the commentbot will now fetch and check user IDs (not usernames) against this file. There is a configurable cache time on reloading the TOML file to avoid spamming the GitHub API and reduce latency on registration calls.

Should fail gracefully allowing all registrations through.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -30,10 +30,18 @@ jobs:
             os: 'ubuntu-latest'
     steps:
       - uses: actions/checkout@v4
+        with:
+          persist-credentials: false
       - uses: julia-actions/setup-julia@v2
         with:
           version: ${{ matrix.version }}
       - uses: julia-actions/julia-buildpkg@latest
+      - name: Generate a token that has read access to https://github.com/JuliaRegistries/user-blocklist-mock-for-testing
+        id: generate-token
+        uses: actions/create-github-app-token@v2
+        with:
+          app-id: ${{ vars.APP_ID_TEST_APP_USER_BLOCKLIST_MOCK }}
+          private-key: ${{ secrets.APP_PRIVATE_KEY_TEST_APP_USER_BLOCKLIST_MOCK }}
       - name: Run the package tests
         run: |
           import Pkg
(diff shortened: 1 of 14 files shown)


DilumAluthge:
Can you also make a PR to change https://github.com/JuliaRegistries/user-blocklist to reflect:
1. TOML instead of plaintext
2. User IDs instead of usernames

rayegun:
@DilumAluthge should the tests run against the actual blocklist repo? That leaks the existence of it and URL but if we just test with @DilumAluthgeBot I don't think it will leak anything. Plus this PR leaks the url and existence anyway haha.

DilumAluthge:
The existence of the banlist repo (and its URL) are totally fine (IMO) to be public. The only thing I want to keep private is the actual list of users that are banlisted.

In theory there's a chance that a CI job would accidentally leak a username or user ID, so for the tests here, maybe we can set up a public "mock" repo just for testing? What do you think?

DilumAluthge:
If you're only testing against DilumAluthgeBot, that seems fine. I would just worry if e.g. the TOML parser throws an error when trying to parse the whole TOML file, or something like that - in theory that might leak other contents of the TOML file?

rayegun:
Perhaps a second private test repo? Exercising the access to private repo using the token is important I think.

DilumAluthge:
Ah, yes we should exercise the "private repo" codepath.

DilumAluthge:
https://github.com/JuliaRegistries/user-blocklist-mock-for-testing

DilumAluthge:
> to both the bot and the webui

Just because the term "bot" can get a bit overloaded (e.g. confused with the AutoMerge checks), can you edit your PR description `bot` ---> `commentbot`, just to be less ambiguous.

rayegun:
Hmm my token works locally for testing. @DilumAluthge does the CI / commentbot have a token with read access to the mock / real repo respectively

rayegun:
Superseded by #481

PR #481, opened 2026-04-03
Add blocklist functionality to commentbot and WebUI

### LINKED ISSUE #443: add mechanism to block users

rayegun:
Heavily vibe-coded by Claude Opus 4.6, but examined line by line manually.

Intended to resolve https://github.com/JuliaRegistries/Registrator.jl/issues/443 by adding a simple configurable blocklist check to both the commentbot and the webui. The blocklist is a (private) repo with a TOML file, and the commentbot will now fetch and check user IDs (not usernames) against this file. There is a configurable cache time on reloading the TOML file to avoid spamming the GitHub API and reduce latency on registration calls.

Should fail gracefully allowing all registrations through.

Replaces #480 in tree.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -30,10 +30,18 @@ jobs:
             os: 'ubuntu-latest'
     steps:
       - uses: actions/checkout@v4
+        with:
+          persist-credentials: false
       - uses: julia-actions/setup-julia@v2
         with:
           version: ${{ matrix.version }}
       - uses: julia-actions/julia-buildpkg@latest
+      - name: Generate a token that has read access to https://github.com/JuliaRegistries/user-blocklist-mock-for-testing
+        id: generate-token
+        uses: actions/create-github-app-token@v2
+        with:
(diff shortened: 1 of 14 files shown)


rayegun:
Alright should be ready. You'll need to add the correct configs to the app and I suppose the WebUI, they are documented in the existing config files.

Can you test with your bot once it's merged?

DilumAluthge:
> Can you test with your bot once it's merged?

It won't take effect immediately after this PR is merged. We'll first need to register a new release of Registrator.jl. But after that, we need to ping the appropriate folks in the JuliaLang Slack[^1], and then those folks would update the production instance of JuliaRegistrator.

[^1]: I usually ping Luke in the `#pkg-registration` channel of the JuliaLang Slack.

DilumAluthge:
CC: @nkottary @lamdor So that you know what's coming down the pipeline.

nkottary (approved):
LGTM

PR #484, opened 2026-04-10
🤖 fix(commentbot): comment when Project.toml fails to parse

### LINKED ISSUE #483: No response from registrator bot if Project.toml is not parseable
I struggled recently to register a new version of a package, getting [no responses](https://github.com/KristianHolme/Drill.jl/commit/45516465d21ff6b736c783788cb11894c4089f29) from the registrator bot. I tried uninstalling and reinstalling the app, commenting on both commits and issues, but nothing worked. 
Then i discovered that i had accidentally duplicated the sources section of the Project.toml, so the file would not parse.

It would have been very helpful if the registrator bot would have picked this up and commented accordingly. It's a basic mistake on my part, it but would have saved me many hours if the bot could have given me a heads up :)

KristianHolme:
- Handle TOML.ParserError and CompositeException/CapturedException
- Include parser showerror text in the GitHub reply message
- Fix RegistryTools.Project error reporting (use showerror, not broken e.msg)
- Wrap ProcessedParams in action() so preprocessing errors still notify users
- Add is_pfile_parseable unit tests

Made-with: Cursor

This hopefully fixes #483.
Most tests pass locally, but i have not tested if the problem from #483 is actually fixed. 
🤖 Code written by Cursor, but looks reasonable to me.

### DIFF
--- a/src/commentbot/CommentBot.jl
+++ b/src/commentbot/CommentBot.jl
@@ -171,17 +171,7 @@ function action(rp::RequestParams{T}, zsock::RequestSocket) where T <: RegisterT
         end
     end
 
-    local pp
-    try
-        pp = ProcessedParams(rp)
-    catch ex
-        @info "Unexpected error while preprocessing registration" exception = (ex, catch_backtrace())
-        msg = "Error while trying to register: preprocessing failed — $(sprint(showerror, ex))"
-        make_comment(rp.evt, msg)
-        set_error_status(rp)
-        @info("Done processing register event", reponame=rp.reponame, target_registry_name)
-        return nothing
-    end
+    pp = ProcessedParams(rp)
     @info("Processing register event", reponame=rp.reponame, target_registry_name)
     try
         if pp.cparams.isvalid
--- a/src/commentbot/CommentBot.jl
+++ b/src/commentbot/CommentBot.jl
@@ -202,9 +202,9 @@ function action(rp::RequestParams{T}, zsock::RequestSocket) where T <: RegisterT
                 set_success_status(rp)
             end
         else
-            @info("Error while processing event: $(repr(pp.cparams.error))")
+            @info("Error while processing event: $(pp.cparams.error)")
             if pp.cparams.report_error
-                msg = "Error while trying to register: $(repr(pp.cparams.error))"
+                msg = "Error while trying to register: $(pp.cparams.error)"
                 @debug(msg)
                 make_comment(rp.evt, msg)
             end
--- a/src/commentbot/CommentBot.jl
+++ b/src/commentbot/CommentBot.jl
@@ -171,7 +171,17 @@ function action(rp::RequestParams{T}, zsock::RequestSocket) where T <: RegisterT
         end
     end
 
-    pp = ProcessedParams(rp)
+    local pp
+    try
+        pp = ProcessedParams(rp)
+    catch ex
+        @info "Unexpected error while preprocessing registration" exception = (ex, catch_backtrace())
+        msg = "Error while trying to register: preprocessing failed — $(sprint(showerror, ex))"
+        make_comment(rp.evt, msg)
+        set_error_status(rp)
+        @info("Done processing register event", reponame=rp.reponame, target_registry_name)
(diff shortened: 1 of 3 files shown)


src/commentbot/verify_projectfile.jl
```diff
@@ -86,10 +111,7 @@ function verify_projectfile_from_sha(reponame, commit_sha; auth=GitHub.Anonymous
                 try
                     project = RegistryTools.Project(TOML.parse(projectfile_contents))
                 catch ex
-                    err = "Failed to read project file"
-                    if isdefined(ex, :msg)
-                        err = err * ": $(e.msg)"
-                    end
+                    err = "Failed to read project file: " * sprint(showerror, ex)
```
nkottary: Would be good if you could beautify this message with some triple back ticks and unescaped newlines. It appears like this currently:

Image

src/commentbot/CommentBot.jl
```diff
@@ -171,7 +171,17 @@ function action(rp::RequestParams{T}, zsock::RequestSocket) where T <: RegisterT
         end
     end
 
-    pp = ProcessedParams(rp)
+    local pp
+    try
+        pp = ProcessedParams(rp)
```
nkottary: This catch is unnecessary. Reporting is already handled below. Please undo all changes to this file.

nkottary (changes requested):
Needs couple of changes otherwise looks good.

src/commentbot/verify_projectfile.jl
```diff
@@ -1,20 +1,45 @@
...
-                @debug(err)
-                return false, err
-            else
-                rethrow(ex)
+            msg = _format_toml_parse_error(ex)
+            if msg !== nothing
+                @debug(msg)
+                return false, msg
             end
+            rethrow(ex)
```
ericphanson: in catch-blocks, prefer just `rethrow()` which will propograte the original exception stack. If you `rethrow(ex)` it starts a new nested exception stack pointing to the catch block. See [here](https://docs.julialang.org/en/v1/base/base/#Base.rethrow)
```suggestion
            rethrow()
```

src/commentbot/verify_projectfile.jl
KristianHolme: Tried to fix this, but have not checked that a real comment looks good.

Before:
```julia
julia> println(err)
Could not parse (Julia)Project.toml as TOML: TOML Parser error:
none:84:-1 error: key already defined
  [sources.DrillInterface]
                         ^
```

After:

````julia
julia> println(err)
Could not parse (Julia)Project.toml as TOML:

```
TOML Parser error:
none:84:-1 error: key already defined
  [sources.DrillInterface]
                         ^
```
````

src/commentbot/CommentBot.jl
KristianHolme: Done!

src/commentbot/verify_projectfile.jl
KristianHolme: Thank you! This is fixed now.

src/commentbot/verify_projectfile.jl
nkottary: The newlines are still escaped and the reason is the use of [`repr`](https://github.com/JuliaRegistries/Registrator.jl/blob/b6626c38c21b0a421037a6975a395a63b2a4d786/src/commentbot/CommentBot.jl#L207). Can you replace that with `something()`?

nkottary (approved):
LGTM!

src/commentbot/verify_projectfile.jl
KristianHolme: I removed `repr`, so now `action` matches `comment_handler`, but did not replace with `something`. 
Did you mean just directly replace like `$(something(pp.cparams.error))`, or something else like `$(something(pp.cparams.error, "" ))`?

nkottary: I meant `$(something(pp.cparams.error, "" ))` but yeah just removing `repr` is correct.

PR #488, opened 2026-06-29
feat: Http2 compat

nkottary:
Support both HTTP.jl 1.x and 2.x in deps

Updates GitHub.jl and GitForge.jl versions as well. These are the HTTP.jl 2.x ported versions.

Removes Mux.jl dependency.

Issue #456, opened 2025-06-20
Is the checked-in `Manifest.toml` file used in any active deployments?


DilumAluthge:
This repo has a `Manifest.toml` file checked in to source control: https://github.com/JuliaRegistries/Registrator.jl/blob/master/Manifest.toml

The manifest file has not been updated in 3 years.

Is this manifest file used in any active deployments of Registrator?

DilumAluthge:
cc: @tanmaykm @nkottary

DilumAluthge:
I'll work on a PR.

PR #388, opened 2022-08-27
Re-resolve and update the `Manifest.toml` file; continue using Julia 1.7.2 for the manifest file

DilumAluthge:
Fixes #389

PR #490, opened 2026-09-05
CI: replace unmaintained Julia 1.7 jobs with 'lts'

eschnett:
CI currently tests on Julia `1.7`, `1`, `nightly`, plus a pinned `1.7.2` job. This replaces the 1.7 jobs with `lts`.

### Why 1.7 should go

- **It has been unmaintained for four years.** Julia maintains exactly two release lines: current stable (1.12) and current LTS (1.10). Julia 1.7.0 was released 2021-11-30, its final patch 1.7.3 on 2022-05-25, and it went unmaintained on 2022-08-17 when 1.8.0 shipped. There is no 1.7.4.

- **The pinned `1.7.2` job no longer has a reason to exist.** It was added in #388 so that one job would match the `julia_version` recorded in the checked-in `Manifest.toml`. That manifest was deleted in #458 (after it was confirmed in #456 that it is not used in deployments), so the job now only shadows the `'1.7'` job.

- **The old jobs were already being tested more weakly than the rest.** `force_use_of_manifest` has been a hardcoded `false` since #458, and #473 added a `VERSION >= v"1.9"` guard so that `force_latest_compatible_version` would not run on 1.7/1.8. With the old jobs removed, both branches collapse into a single unconditional `Pkg.test` call.

- **1.7 was silently giving no coverage of the newest code.** `HTTP` v2 declares `julia = "1.10"` in the General registry, so the HTTP2 support added in #488 could never be installed on the 1.7 jobs — they resolved `HTTP` v1 no matter what. Same for `GitForge` 0.4.5.

### Second commit (separable)

`[compat] julia = "1.1"` has never been checked by CI and is not accurate: `GitHub` 5.11+, `RegistryTools` 2.4+, `GitForge` 0.4.3+ and `URIs` 1.4+ all require Julia ≥ 1.6, and `HTTP` v2 requires ≥ 1.10. The second commit sets it to `1.10` so the declared bound matches the oldest version CI runs, and drops the now-dead `Base.VERSION >= v"1.7-"` guard in `test/util.jl`.

**I have deliberately kept this as its own commit.** I could not find any public record of which Julia version the deployed `@JuliaRegistrator` instance runs on — `Dockerfile` says `FROM julia:latest`, but that is the self-hosting reference image, not the deployment. If production is on something older than 1.10, please drop the second commit and take only the CI change; the two are independent.

Happy to adjust either way.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -20,14 +20,14 @@ jobs:
       fail-fast: false
       matrix:
         version:
-          - '1.7'
+          # Julia maintains only two release lines: the current stable release
+          # and the current long-term support (LTS) release. Everything else is
+          # unmaintained.
+          - 'lts'
           - '1'
           - 'nightly'
         os:
           - ubuntu-latest
-        include:
-          - version: '1.7.2'
-            os: 'ubuntu-latest'
     steps:
       - uses: actions/checkout@v4
       - uses: julia-actions/setup-julia@v2
@@ -37,15 +37,10 @@ jobs:
       - name: Run the package tests
         run: |
           import Pkg
-          force_use_of_manifest = false
-          allow_reresolve = !force_use_of_manifest
+          allow_reresolve = true
           coverage = true
-          if allow_reresolve && VERSION >= v"1.9"
-            force_latest_compatible_version = true
-            Pkg.test(; allow_reresolve, coverage, force_latest_compatible_version)
-          else
-            Pkg.test(; allow_reresolve, coverage)
-          end
+          force_latest_compatible_version = true
+          Pkg.test(; allow_reresolve, coverage, force_latest_compatible_version)
         env:
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
         shell: julia --color=yes --project {0}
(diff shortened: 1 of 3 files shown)


eschnett:
The three failing jobs here are **not caused by this PR**. `master` is red for an unrelated reason, and this PR inherits it.

- All three test jobs fail, including `1` and `nightly`, whose matrix entries and resolution behavior this PR leaves byte-identical. Only the `1.7` and `1.7.2` rows changed.
- The last CI run on `master` was 2026-07-01 (51847d8, green). It has not run since.
- The cause is **URIs.jl v1.7.0**, released 2026-08-13, which tightened URI validation. `URI("https://foo:bar\">&baz@github.com/owner/repo")` — built by `test/webui/gitutils.jl` — now throws `ParseError("Invalid URI userinfo: ...")`, because `"` and `>` are not legal literals in userinfo under RFC 3986 §3.2.1. URIs v1.6.3, four days earlier, accepted it.

I have opened #491 with the fix. Once that merges I will rebase this branch, and these jobs should go green.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

